PR #79, opened 2018-05-04
Issue #53: Cleanup repeated functionality in code

### LINKED ISSUE #53: Cleanup repeated functionality in code

alexandermushinskiy:
Fixes #53 

Changes:
- remove unused legacy code in wallet service;
- remove network service;
- updates in wallet-detail component:
a) ability to copy address to clipboard by clicking on it;
b) open QR code in a modal window by clicking on it.


src/app/services/wallet.service.ts
```diff
@@ -176,53 +173,11 @@ export class WalletService {
     return this.apiService.get('pendingTxs');
   }
 
-  recent(): Observable<any[]> {
-    return this.recentTransactions.asObservable();
-  }
-
-  refreshBalances() {
```
Senyoret1: I think that `refreshBalances` should not have been deleted. This is due to two things:

- The desktop wallet uses `refreshBalances`. Using refreshBalances would help to maintain the consistency between the two wallets. At this time the web wallet uses `loadBalances` instead.

- `loadBalances` seems to have several important problems, such as that it can send too many parameters through the URL and throw errors because of that.

A more in-depth review should be done to decide which of the two functions will be used at the end (probably with modifications). Also `retrieveAddressBalance` and `retrieveWalletBalance` would not be needed if `refreshBalances` is deleted.

alexandermushinskiy:
@Senyoret1, can we merge this PR and close or there are other comments?

Senyoret1:
Wow, I made a silly mistake in my last comment, sorry. I've corrected it. Only what I said in my last comment remains pending.

PR #92, opened 2018-05-08
Issue #86: Coins validation issues

### LINKED ISSUE #86: Coins validation issues
There are problems in the validation of the amount of coins in the send coins form. To solve it, please consult this pr and the associated issue: https://github.com/skycoin/skycoin/pull/1350.

alexandermushinskiy:
Fixes #86 

Changes:
- in send coin page: change validator for amount value, update reset form values
- new directive was created to allow specify only numbers

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -67,6 +67,7 @@ import { AppRoutes } from './app.routes';
 import { CipherProvider } from './services/cipher.provider';
 import { FeatureToggleModule } from 'ngx-feature-toggle';
 import { FeatureService } from './services/feature.service';
+import { NumberFieldDirective } from './directives/number-field.directive';
 
 @NgModule({
   declarations: [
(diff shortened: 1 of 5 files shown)


alexandermushinskiy:
We switched from `type="number"` to `type="text"`, because number control can't validate numbers with big amount of digits as you mentioned in your previous post: 
> accepts values with more decimals than the maximum allowed (like 20.500000000000000001)
it's been rounded to 20 and that is why it validates as correct value.

alexandermushinskiy:
As a proposition, we can switch back to `type="number"` and disable ability to specify more than 6 digits after decimal. What do you think?

Senyoret1:
Let's do something: switch it back to `type="number"`, remove `NumberFieldDirective` from the input, and add a comment at the top of number-field.directive.ts with something like "Allows an input to accept only numbers. Currently, the paste option is failing, it needs some more testing". That way we keep the code for `NumberFieldDirective` (it would be a pity to delete it) and the final implementation could be reviewed with more detail in another pr, so I can merge this pr with more confidence. `validateAmount` is already responsible for preventing the user from continuing if the amount have more than 6 decimals.

alexandermushinskiy:
@Senyoret1, we have switched back to `type="number"`.
Also have we updated copy-paste logic of NumberFieldDirective: it allows paste only if clipboard contains a number (can be decimal with dot), otherwise (if there is space, comma, letter, etc.) it prevents paste.

PR #100, opened 2018-05-09
Issue #69: Problems in the process for creating transactions

### LINKED ISSUE #69: Problems in the process for creating transactions
The way in which `WalletService.sendSkycoin` work creates excessively large transactions and a greater spending of coin hours than necessary. For each transaction, all the outputs are being used, instead of making a calculation about the outputs that are necessary. Additionally, the calculation causes hours with decimals to be sent, which is not allowed. As a reference, the node function that creates the transactions for the desktop wallet is `src/wallet/wallet.go/CreateAndSignTransactionAdvanced` (https://github.com/skycoin/skycoin/blob/develop/src/wallet/wallet.go#L1026).

AndreiIarovoi:
Fixes #69 

Changes:
- getMinRequiredOutputs method has been created in the wallet.service.ts to retrieve minimum required outputs to create transaction.
- Minimum required outputs are used for both transaction inputs and outputs.
- wallet.service.spec.ts unit tests has been adjusted based on the above described changes.


Senyoret1:
There are some problems, such as that the outputs are being sorted by hours and not by coins (so `getMinRequiredOutputs` does not really return as few outputs as possible) and that it would be possible to create a transaction without hours to burn.

However, the problem is even more complicated than what I had anticipated. There are special cases in which an additional output is used to make the user lose less hours, as explained here: https://github.com/skycoin/skycoin/issues/1428. Likewise, there are plans for some rules to change over time. Taking all this into account, I think it is best to wait for the result of this: https://github.com/skycoin/skycoin/issues/1461

gz-c:
@Senyoret1 we can use a simple solution here for now, even if it is not as optimal

src/app/services/wallet.service.ts
```diff
@@ -57,19 +59,22 @@ export class WalletService {
...
-      const totalCoins = parseInt((outputs.reduce((count, output) => count + output.coins, 0) * 1000000) + '', 10);
-      const totalHours = outputs.reduce((count, output) => count + output.hours, 0);
+      const minRequiredOutputs =  this.getMinRequiredOutputs(amount, outputs);
+      const totalCoins = parseInt((minRequiredOutputs.reduce((count, output) =>
+        count + output.coins, 0) * 1000000) + '', 10);
+      const totalHours = minRequiredOutputs.reduce((count, output) => count + output.hours, 0);
       const changeCoins = totalCoins - amount * 1000000;
-      const changeHours = totalHours / 4;
+      const changeHours = Math.floor(totalHours * this.allocationRatio);
       const txOutputs: TransactionOutput[] = [{ address: address, coins: amount * 1000000, hours: changeHours }];
```
Senyoret1: Please use `parseInt` with the coins. Same reason as with line 68.

Also, if `changeCoins == 0` then `hours` should be `changeHours * 2`, filtered with `parseInt`. This is because, if it is not possible to return the remaining hours to the user, they should be sent to the recipient.

src/app/services/wallet.service.ts
```diff
@@ -318,4 +323,22 @@ export class WalletService {
...
+  private getMinRequiredOutputs(amount: number, outputs: Output[]): Output[] {
+    outputs.sort( function(a, b) {
+      return a.hours - b.hours;
+    });
+
+    const minRequiredOutputs: Output[] = [];
+    let sumCoins = 0;
+
+    outputs.forEach(output => {
+      if (amount > sumCoins) {
```
Senyoret1: Any output without hours must be ignored. This, however, brings an additional problem: it is possible that the usable outputs do not have enough coins to complete the amount that the user wants to send. In this case, the entire process of sending coins must fail, in such a way that it is possible to inform the user the reason of the failure.

src/app/services/wallet.service.ts
```diff
@@ -57,19 +59,22 @@ export class WalletService {
...
   sendSkycoin(wallet: Wallet, address: string, amount: number) {
     const addresses = wallet.addresses.map(a => a.address).join(',');
+
     return this.apiService.getOutputs(addresses).flatMap((outputs: Output[]) => {
-      const totalCoins = parseInt((outputs.reduce((count, output) => count + output.coins, 0) * 1000000) + '', 10);
-      const totalHours = outputs.reduce((count, output) => count + output.hours, 0);
+      const minRequiredOutputs =  this.getMinRequiredOutputs(amount, outputs);
+      const totalCoins = parseInt((minRequiredOutputs.reduce((count, output) =>
+        count + output.coins, 0) * 1000000) + '', 10);
+      const totalHours = minRequiredOutputs.reduce((count, output) => count + output.hours, 0);
```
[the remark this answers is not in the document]
AndreiIarovoi: As far as I can see, ApiService.getOutputs already returns "calculated_hours" hours in the hours field.
Do we need expand the "Output" model to have both hours and calculated_hours or rename "hours" field to the "calculated_hours" to have a clear naming?

src/app/services/wallet.service.ts
AndreiIarovoi: Can we remove outputs without hours from the minRequiredOutputs array and calculate total outputs' amount before sending transaction? In that case we can send toaster "Not enough SKY Hours available to perform transaction" notification in case total outputs coins is less then amount that user wants to send.

src/app/services/wallet.service.ts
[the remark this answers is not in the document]
AndreiIarovoi: Agree, please add a new issue and we wail take care about that.

src/app/services/wallet.service.ts
Senyoret1: Sounds good, but I guess you have to remove them from the `outputs` array, instead of `minRequiredOutputs`.

src/app/services/wallet.service.ts
```diff
@@ -57,19 +59,22 @@ export class WalletService {
...
       const changeCoins = totalCoins - amount * 1000000;
```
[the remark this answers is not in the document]
AndreiIarovoi: Fixed

src/app/services/wallet.service.ts
AndreiIarovoi: Do you mean that all outputs retrieved by apiService.getOutputs() except the ones with empty sky hours should be used as inputs for the current transaction (currently outputs retrieved by getMinRequiredOutputs() method are used)?

Senyoret1: No, It is ok as you have thought: to use only the smallest possible amount of outputs. Maybe I misunderstood your previous comment, but what I was saying is that you should probably remove the outputs without hours from the `outputs` array before entering the foreach, because if you remove the outpus without hours from the `minRequiredOutputs` array after the loop, errors are very likely to occur. However, if I misunderstood you comment, please proceed with what you have in mind, I am supposed to check it out again later anyway.

AndreiIarovoi: Thank you for the prompt answers.

src/app/services/wallet.service.ts
AndreiIarovoi: In case we have a wallet with coins: 1, hours: 10 and want to send 1 coin.
We need to have one output with coins: 1 and hours: ...
a) 10/4 = 2.5 => 2, and since there is no change => 2*2 = 4
b) 10/2 = 5 => 5
Could you, please, clarify what is the correct answer (4 or 5)?

Senyoret1: The correct answer is 5. In fact, if the example had an output with 2 coins, instead of 1, and the user wanted to send just 1 coin, the most appropriate thing would be to send 2 hours to the recipient and return 3 to the user.

src/app/services/wallet.service.ts
```diff
@@ -64,16 +65,31 @@ export class WalletService {
       const minRequiredOutputs =  this.getMinRequiredOutputs(amount, outputs);
       const totalCoins = parseInt((minRequiredOutputs.reduce((count, output) =>
         count + output.coins, 0) * 1000000) + '', 10);
-      const totalHours = minRequiredOutputs.reduce((count, output) => count + output.hours, 0);
+
+      if (totalCoins < amount * 1000000) {
```
Senyoret1: The result of the multiplication must be filtered with `parseInt`, to prevent errors.

src/app/services/wallet.service.ts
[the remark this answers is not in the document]
Senyoret1: The result of the multiplication still must be filtered with `parseInt`, to prevent errors.

src/app/services/wallet.service.ts
```diff
@@ -64,16 +65,31 @@ export class WalletService {
...
+        txOutputs.push({
+          address: wallet.addresses[0].address,
+          coins: changeCoins,
+          hours: totalHours * this.unburnedHouarsRatio - changeHours
+        });
+      } else {
+        changeHours = parseInt((totalHours * this.unburnedHouarsRatio) + '', 10);
       }
 
+      txOutputs.push({ address: address, coins: amount * 1000000, hours: changeHours });
```
Senyoret1: The result of the multiplication must be filtered with `parseInt`, to prevent errors.

src/app/services/wallet.service.ts
```diff
@@ -64,16 +65,31 @@ export class WalletService {
...
+      if (totalCoins < amount * 1000000) {
+        throw new Error('Not enough available SKY Hours to perform transaction!');
+      }
+
+      const totalHours = parseInt((minRequiredOutputs.reduce((count, output) =>
+        count + output.hours, 0)) + '', 10);
       const changeCoins = totalCoins - amount * 1000000;
-      const changeHours = Math.floor(totalHours * this.allocationRatio);
-      const txOutputs: TransactionOutput[] = [{ address: address, coins: amount * 1000000, hours: changeHours }];
+      let changeHours = parseInt((totalHours * this.allocationRatio) + '', 10);
```
Senyoret1: Now this variable is used in another way, it does not longer store the number of change hours, but the amount of hours that will be sent to the destination address, so the name should be updated to something like "hoursToSend".

Senyoret1:
The second comment of the last review (line 63/68) was marked as outdated, but it is still valid.

src/app/services/wallet.service.ts
AndreiIarovoi: All comments are fixed.

PR #101, opened 2018-05-09
 #99 - The texts of the seeds must be cleaned

### LINKED ISSUE #99: The texts of the seeds must be cleaned
It is necessary to clean the unwanted characters of the seeds. Please see https://github.com/skycoin/skycoin/blob/develop/src/gui/static/src/app/services/wallet.service.ts#L54

AndreiIarovoi:
Fixes #99 

Changes:
- Clean the unwanted characters of the seeds by trimming a trailing newline for the create and unlock 
wallet functionality.


Senyoret1:
It works correctly, but some changes were requested. I think it would be better to make the changes after they are added to the desktop wallet: https://github.com/skycoin/skycoin/issues/1459

PR #106, opened 2018-05-10
Issue #96: The wallet is not periodically updated

### LINKED ISSUE #96: The wallet is not periodically updated
Due to this problem, if the user receives coins while the wallet is open, he will not be able to see the updated balance until the browser is manually refreshed. Like the desktop wallet, the balance of the web wallet could be updated every 30 seconds, but some users will leave the wallet open for a long period of time without use, which would cause unnecessary bandwidth usage.

An alternative would be to refresh only every 5 minutes and add a text indicating since how many minutes the information was updated. The text could look like this:
time-updated

Another alternative would be to refresh the balance every 30 seconds while the browser window is active and every 10 minutes while it is not. This option is a bit more complicated because the best method could have compatibility problems if the browser is not updated: https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API#Browser_compatibility . Or we could simply use the `window.onfocus` event, but it reports `true` when the window is active, regardless of whether the user is using another tab.

What do you think of this, @gz-c ?

alexandermushinskiy:
Fixes #96 

Changes:
- Updates wallets balances each 5 minutes;
- Show time (in a minutes) how long ago balances were updated;


src/app/services/wallet.service.ts
```diff
@@ -35,6 +37,10 @@ export class WalletService {
     return this.wallets.asObservable().map(wallets => wallets ? wallets : []);
   }
 
+  getBalancesUpdated(): Observable<Date> {
```
Senyoret1: This does not follow the style that has been used in `WalletService`. It would be better if it changed to `get balancesUpdateDate()`

src/app/components/layout/header/top-bar/top-bar.component.ts
```diff
@@ -1,10 +1,60 @@
...
 })
-export class TopBarComponent {
+export class TopBarComponent implements OnInit, OnDestroy {
   @Input() title: string;
+
+  timeSinceLastUpdateBalances = 0;
+
+  private updateBalancesTimer: NodeJS.Timer;
+  private updateBalancesSubscription: Subscription;
+  private lastUpdatedBalancesTime;
```
Senyoret1: `lastBalancesUpdateTime` seems more apropiate.

src/app/components/layout/header/top-bar/top-bar.component.ts
```diff
@@ -1,10 +1,60 @@
...
+  private intervalTime = 60 * 1000;
+  private refreshBalancesTime = 5;
+
+  constructor(private walletService: WalletService) {
+  }
+
+  ngOnInit() {
+    this.updateBalancesSubscription = this.walletService.getBalancesUpdated()
+      .subscribe((lastUpdatedTime) => {
```
Senyoret1: `lastUpdateTime` seems more apropiate.

alexandermushinskiy:
@Senyoret1, please be informed that your review comments are fixed.

alexandermushinskiy:
@Senyoret1, please be informed that rest of review comments are fixed.

src/app/services/wallet.service.ts
```diff
@@ -18,6 +19,13 @@ export class WalletService {
   recentTransactions: Subject<any[]> = new BehaviorSubject<any[]>([]);
   wallets: BehaviorSubject<Wallet[]> = new BehaviorSubject<Wallet[]>([]);
   addressesTemp: Address[];
+  timeSinceLastBalancesUpdate: BehaviorSubject<number> = new BehaviorSubject<number>(null);
+  totalBalance: BehaviorSubject<TotalBalance> = new BehaviorSubject<TotalBalance>(null);
+
+  private updateBalancesTimer: any;
+  private lastBalancesUpdateTime: Date;
+  private readonly intervalTime = 60 * 1000;
+  private readonly refreshBalancesTime = 2;
```
Senyoret1: This must be 5.

src/app/services/wallet.service.ts
```diff
@@ -200,10 +208,17 @@ export class WalletService {
       .reduce((a , b) => a + b, 0));
   }
 
-  private addWallet(wallet) {
-    this.all.first().subscribe(wallets => {
-      wallets.push(wallet);
-      this.saveWallets(wallets);
+  transaction(txid: string): Observable<any> {
```
Senyoret1: This function does not seem to be used anywhere.

e2e/onboarding-create.e2e-spec.ts
```diff
@@ -1,9 +1,12 @@
+import { browser } from 'protractor';
+
 import { OnboardingCreatePage } from './onboarding-create.po';
 
 describe('Onboarding Create', () => {
   let page: OnboardingCreatePage;
 
   beforeEach(() => {
+    browser.waitForAngularEnabled(false);
```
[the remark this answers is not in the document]
alexandermushinskiy: Without adding browser.waitForAngularEnabled(false) all our e2e tests are failed because setInterval() function is used to calculate last updated time.
Due to it is used in the header component it is affected all pages.

Why is this changes being made in several test files? - 'waitForAngularEnabled' is a configuration of browser and it's more than enough to add it in the first test file (onboarding-create.e2e-spec.ts), just because it is run firstly. We decided to added it to each test file just to make them independent from each other. We will remove 'waitForAngularEnabled' from all files except onboarding-create.e2e-spec.ts

src/app/components/layout/header/header.component.ts
```diff
@@ -10,36 +12,51 @@ import { WalletService } from '../../../services/wallet.service';
 })
 export class HeaderComponent implements OnInit, OnDestroy {
   @Input() title: string;
-  @Input() coins: number;
-  @Input() hours: number;
+
+  coins = 0;
+  hours: number;
+  balance: string;
+  isBalanceCalculated = false;
```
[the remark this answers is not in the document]
alexandermushinskiy: Fixed

src/app/services/wallet.service.ts
alexandermushinskiy: Fixed

src/app/services/wallet.service.ts
alexandermushinskiy: Fixed

PR #113, opened 2018-05-14
Issues #94: It should not be assumed that the node is available and synchronized

### LINKED ISSUE #94: It should not be assumed that the node is available and synchronized.
At some point, the node to which the wallet connects could have a problem, so it may be inaccessible or desynchronized. At this time, the wallet simply shows balance 0, which can cause users to panic and does not provide any information about the existence of a problem.

To improve the user experience, I think the wallet should show the percentage of synchronization of the node, in case it is not fully synchronized, just like the desktop wallet does. If the node is inaccessible, a red alert box should be placed at the bottom of the header with a text like "No connection to the Skycoin node", and the balance text should say "Loading".

A good starting point for these changes is the code of the HeaderComponent component of the desktop wallet: https://github.com/skycoin/skycoin/tree/develop/src/gui/static/src/app/components/layout/header

alexandermushinskiy:
Fixes #94 

Changes:
- added app.service to check connection state;
- added ConnectionError enumeration to identify a connection error;
- added functionality of loading blockchain blocks (syncing) and show the progress in UI;


src/app/services/wallet.service.ts
```diff
@@ -180,14 +180,21 @@ export class WalletService {
       .reduce((a , b) => a + b, 0));
   }
 
-  private addWallet(wallet) {
-    this.all.first().subscribe(wallets => {
-      wallets.push(wallet);
-      this.saveWallets(wallets);
+  transaction(txid: string): Observable<any> {
```
Senyoret1: This function apparently it is not used anywhere. Is there any particular reason why it was added?

src/app/services/blockchain.service.ts
```diff
@@ -44,11 +57,32 @@ export class BlockchainService {
     return this.blocks(1).map(blocks => blocks[0]);
   }
 
-  progress() {
+  private retrieveInputAddress(input: string) {
+    return this.apiService.get('uxout', {uxid: input});
+  }
+
+  private loadBlockchainBlocks() {
+    setTimeout(() => IntervalObservable
```
Senyoret1: I did some testing and it seems that the procedure works correctly without needing to wait 3 seconds. Is there a particular case in which this is necessary?

src/app/services/blockchain.service.ts
```diff
@@ -44,11 +57,32 @@ export class BlockchainService {
...
+      .create(2000)
```
Senyoret1: Please change this to 90000.

src/app/services/blockchain.service.ts
alexandermushinskiy: It was copied from the desktop wallet. Will be removed in next commit.

src/app/services/wallet.service.ts
alexandermushinskiy: No particular reason. Will be removed in next commit.

src/app/services/blockchain.service.ts
alexandermushinskiy: Will be changed in next commit.

src/app/services/blockchain.service.ts
```diff
@@ -62,15 +62,15 @@ export class BlockchainService {
   }
 
   private loadBlockchainBlocks() {
-    setTimeout(() => IntervalObservable
-      .create(2000)
+    IntervalObservable
+      .create(90000)
       .flatMap(() => this.getBlockchainProgress())
```
Senyoret1: This needs an additional change. As IntervalObservable does not run for the first time immediately, but after the established time has elapsed, it is now necessary to wait 3 minutes until the wallet shows the balance. This should not happen, the connection must be made immediately. One option is to put `.startWith(1)` just below `.create(90000)`.

Senyoret1:
Now there are several conflicts.

alexandermushinskiy:
Comments are fixed, conflicts are resolved.

Issue #122, opened 2018-05-16
Use /balance instead of /outputs in WalletService.loadBalances


Senyoret1:
This change will reduce the amount of information that the node must return.

PR #126, opened 2018-05-16
Issue #123: Show an alert when there are pending transactions

### LINKED ISSUE #123: Show an alert when there are pending transactions
Each time the balance is updated, the list of pending transactions must be verified. If there are any pending transactions affecting any of the user addresses, an alert should be shown to the user, just as happens in the desktop wallet right after making a transaction.

Additionally, while the alert is displayed the balance update period should be reduced from 5 minutes to 20 seconds.

alexandermushinskiy:
Fixes #123 

Changes:
- show notification bar if there is at least one pending transaction.

### DIFF
--- a/src/app/services/wallet.service.ts
+++ b/src/app/services/wallet.service.ts
@@ -315,10 +315,10 @@ export class WalletService {
   }
 
   private calculateTimeSinceLastUpdate() {
-    const diffMs: number = this.lastBalancesUpdateTime.getTime() - new Date().getTime();
-    const timeSinceLastUpdate = this.convertDecimalToInt(diffMs / 1000);
-
     this._ngZone.run(() => {
+      const diffMs: number = this.lastBalancesUpdateTime.getTime() - new Date().getTime();
+      const timeSinceLastUpdate = this.convertDecimalToInt(diffMs / 1000);
+
       this.timeSinceLastBalancesUpdate.next(this.convertDecimalToInt(timeSinceLastUpdate / 60));
 
       if (timeSinceLastUpdate === this.refreshBalancesTimeInSec) {


Senyoret1:
https://github.com/skycoin/skycoin-web/issues/122 has been solved. Now the balance is updated using this api endpoint: https://github.com/skycoin/skycoin/blob/develop/src/api/README.md#get-balance-of-addresses. The relevant part is that this endpoint returns two balances:

- confirmed: the current confirmed balance of all the addresses.
- predicted: the balance that the addresses will have after the pending transactions are processed.

This means that, if the current balance and the predicted balance do not match, then there are pending transactions affecting one or more of the user's addresses. It is important to keep in mind that it is necessary to check both the coins and the hours, if any of the two values does not match, then there are pending transactions. Knowing this, it is no longer necessary to call `/pendingTxs`, which simplifies the code and saves bandwidth.

alexandermushinskiy:
In the last commit:
- merged with master;
- updated logic to define existing pending transaction using balance properties confirmed and predicted;
- removed method which returns observable pending transactions to make it consistence with whole code;
- added hasPendingTransactions (ReplaySubject<boolean>) instead of pendingTransaction (ReplaySubject<any[]>);
- set update time to 20 seconds instead of 20 minutes.

src/app/services/wallet.service.ts
```diff
@@ -311,17 +316,21 @@ export class WalletService {
 
   private calculateTimeSinceLastUpdate() {
     const diffMs: number = this.lastBalancesUpdateTime.getTime() - new Date().getTime();
-    const timeSinceLastUpdate = Math.abs(Math.round((diffMs / 1000 / 60)));
+    const timeSinceLastUpdate = this.convertDecimalToInt(diffMs / 1000);
```
Senyoret1: `diffMs` and `timeSinceLastUpdate` must be inside `this._ngZone.run(() => {`, otherwise the updating is not done correctly.

alexandermushinskiy: The comment has been fixed.

PR #140, opened 2018-05-20
Add cryptography info to README.md

Senyoret1:
Fixes # non reported

Changes:
- Adds cryptography info to README.md

Please check if the information is adequate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,3 +32,19 @@ Run `npm start` for a dev server. Navigate to `http://localhost:4200/`.
 As this is a lite client, it requires a back-end to retrieve the blockchain state and inject new transactions. For this
 purpose a full node has been set up at `http://128.199.57.221`. At the moment this requires a mapping API, but in the
 future any node can do this.
+
+## Cryptography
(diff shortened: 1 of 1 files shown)


gz-c:
How does the JS store the seed temporarily?

Senyoret1:
@gz-c , at this moment there is an `wallets` array in `WalletService`:
https://github.com/skycoin/skycoin-web/blob/ff2153f719ce122e312d0a24b36ccaeb479bca76/src/app/services/wallet.service.ts#L18

Each element has the following structure:
https://github.com/skycoin/skycoin-web/blob/ff2153f719ce122e312d0a24b36ccaeb479bca76/src/app/app.datatypes.ts#L1-L19

So, if the user has just created the wallet or introduced the seed to do some operation, the seed is saved in the `seed` property until the tab/window is closed or the user refreshes the browser. In addition, if the  `seed` property have a valued, then each address within the wallet object stores the seed necessary for creating the next address, in the `next_seed` property.

PR #156, opened 2018-05-24
Issue #148: Add a transaction verification screen

### LINKED ISSUE #148: Add a transaction verification screen
Please use https://github.com/skycoin/skycoin/pull/1408 as reference. The desktop wallet uses an api endpoint to obtain the transaction data, while the web wallet will have to build the transaction manually (at least for now), as it is doing it in `WalletService.sendSkycoin`.

alexandermushinskiy:
Fixes #148 

Changes:
- splitted up the sendSkycoin method into two separate methods: createTransaction() and injectTransaction();
- added three new components: SendForm, SendVerify, TransactionInfo
- verification page is shown before actual sending of coins;
- added two separate data types: NormalTransaction and PreviewTransaction (both of them extends Transaction);
- 100000 multiplier is used when transaction is going to be posted (not in createTransaction);
- updated balance after transaction has been posted;
- added/updated appropriate unit tests;

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -106,8 +106,7 @@ import { TransactionInfoComponent } from './components/pages/send-skycoin/send-v
     WalletDetailComponent,
     SendFormComponent,
     SendVerifyComponent,
-    TransactionInfoComponent,
-    WalletDetailComponent
+    TransactionInfoComponent
   ],
   entryComponents: [
     AddDepositAddressComponent,
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -66,6 +66,9 @@ import { FeatureToggleModule } from 'ngx-feature-toggle';
 import { FeatureService } from './services/feature.service';
 import { AppService } from './services/app.service';
 import { NumberFieldDirective } from './directives/number-field.directive';
+import { SendFormComponent } from './components/pages/send-skycoin/send-form/send-form.component';
+import { SendVerifyComponent } from './components/pages/send-skycoin/send-verify/send-verify.component';
+import { TransactionInfoComponent } from './components/pages/send-skycoin/send-verify/transaction-info/transaction-info.component';
 
 @NgModule({
   declarations: [
@@ -101,6 +104,9 @@ import { NumberFieldDirective } from './directives/number-field.directive';
     UnlockWalletComponent,
     WalletsComponent,
     WalletDetailComponent,
+    SendFormComponent,
+    SendVerifyComponent,
+    TransactionInfoComponent
   ],
   entryComponents: [
     AddDepositAddressComponent,
--- a/src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
+++ b/src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
@@ -32,8 +32,6 @@ export class SendVerifyComponent {
         () => this.onSuccess(),
         (error) => this.onError(error)
       );
-
-    this.walletService.loadBalances();
   }
 
   back() {
@@ -42,6 +40,9 @@ export class SendVerifyComponent {
 
   private onSuccess() {
     this.sendButton.setSuccess();
+    this.sendButton.setDisabled();
+
+    this.walletService.loadBalances();
     setTimeout(() => this.onBack.emit(true), 3000);
   }
 
--- a/src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
+++ b/src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
@@ -1,0 +1,52 @@
+import { Component, EventEmitter, Input, Output, ViewChild } from '@angular/core';
+import { MatSnackBar, MatSnackBarConfig } from '@angular/material';
+
+import { PreviewTransaction } from '../../../../app.datatypes';
+import { WalletService } from '../../../../services/wallet.service';
+import { ButtonComponent } from '../../../layout/button/button.component';
+
+@Component({
+  selector: 'app-send-verify',
+  templateUrl: './send-verify.component.html',
+  styleUrls: ['./send-verify.component.scss'],
+})
+export class SendVerifyComponent {
+  @ViewChild('sendButton') sendButton: ButtonComponent;
+  @ViewChild('backButton') backButton: ButtonComponent;
+  @Input() transaction: PreviewTransaction;
+  @Output() onBack = new EventEmitter<boolean>();
+
+  constructor(
+    private walletService: WalletService,
+    private snackbar: MatSnackBar
+  ) {}
+
+  send() {
+    this.snackbar.dismiss();
+    this.sendButton.resetState();
+    this.sendButton.setLoading();
+
+    this.walletService.injectTransaction(this.transaction.inputs, this.transaction.outputs)
+      .subscribe(
+        () => this.onSuccess(),
+        (error) => this.onError(error)
+      );
+  }
+
+  back() {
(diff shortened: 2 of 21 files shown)


src/app/services/wallet.service.ts
```diff
@@ -115,6 +116,62 @@ export class WalletService {
     });
   }
 
+  createTransaction(wallet: Wallet, address: string, amount: number): Observable<{ inputs: TransactionInput[], outputs: TransactionOutput[] }> {
+    const addresses = wallet.addresses.map(a => a.address).join(',');
+
+    return this.apiService.getOutputs(addresses)
+      .flatMap((outputs: Output[]) => {
+        const minRequiredOutputs =  this.getMinRequiredOutputs(amount, outputs);
+        const totalCoins = parseInt((minRequiredOutputs.reduce((count, output) => count + output.coins, 0)) + '', 10);
```
Senyoret1: As the multiplication by 1000000 was removed, it is necessary to remove `parseInt`, or decimal values of the user's outputs will be ignored. Instead filter the value with `Number(([operation]).toFixed(6))`.

src/app/services/wallet.service.ts
```diff
@@ -115,6 +116,62 @@ export class WalletService {
...
+        }
+
+        const totalHours = parseInt((minRequiredOutputs.reduce((count, output) => count + output.calculated_hours, 0)) + '', 10);
+        let hoursToSend = parseInt((totalHours * this.allocationRatio) + '', 10);
+
+        const txOutputs: TransactionOutput[] = [];
+        const txInputs: TransactionInput[] = [];
+        const calculatedHours = parseInt((totalHours * this.unburnedHoursRatio) + '', 10);
+
+        const changeCoins = totalCoins - amount;
```
Senyoret1: Please, filter with `Number(([operation]).toFixed(6))`.

src/app/services/wallet.service.ts
```diff
@@ -115,6 +116,62 @@ export class WalletService {
...
+          });
+        });
+
+        return Observable.of({ inputs: txInputs, outputs: txOutputs });
+    });
+  }
+
+  injectTransaction(txInputs: TransactionInput[], txOutputs: TransactionOutput[]): Observable<string> {
+    txOutputs.forEach(output => {
+      output.coins = output.coins * this.coinsMultiplier;
```
Senyoret1: This must be filtered with `parseInt`.

src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
```diff
@@ -0,0 +1,52 @@
...
+  constructor(
+    private walletService: WalletService,
+    private snackbar: MatSnackBar
+  ) {}
+
+  send() {
+    this.snackbar.dismiss();
+    this.sendButton.resetState();
+    this.sendButton.setLoading();
+
```
Senyoret1: `backButton` should be disabled here, to prevent the user from returning while the transaction is being sent.

src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  back() {
+    this.onBack.emit(false);
+  }
+
+  private onSuccess() {
+    this.sendButton.setSuccess();
+    this.walletService.loadBalances();
+    this.onBack.emit(true);
```
Senyoret1: Please, do not return immediately, but after a moment, as in the desktop wallet. That small amount of time, with `sendButton.setSuccess()`, is currently being used as a method for the user to know that the transaction was sended.

Also, refresh the balance just after the transaction is sent.

alexandermushinskiy: Fixed.

src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
alexandermushinskiy: Fixed.

src/app/services/wallet.service.ts
alexandermushinskiy: Fixed.

src/app/services/wallet.service.ts
alexandermushinskiy: Fixed.

src/app/services/wallet.service.ts
alexandermushinskiy: Fixed.

src/app/app.module.ts
```diff
@@ -106,7 +106,8 @@ import { TransactionInfoComponent } from './components/pages/send-skycoin/send-v
     WalletDetailComponent,
     SendFormComponent,
     SendVerifyComponent,
-    TransactionInfoComponent
+    TransactionInfoComponent,
+    WalletDetailComponent
```
Senyoret1: This was already present in line 106.

src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
```diff
@@ -39,8 +42,7 @@ export class SendVerifyComponent {
 
   private onSuccess() {
     this.sendButton.setSuccess();
-    this.walletService.loadBalances();
```
Senyoret1: This was good here. In a previous comment I said that the balance had to be updated just after sending the transaction. This was doing precisely that, but I did not notice.

src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
```diff
@@ -25,12 +25,15 @@ export class SendVerifyComponent {
...
     this.sendButton.setLoading();
+    this.backButton.setDisabled();
 
     this.walletService.injectTransaction(this.transaction.inputs, this.transaction.outputs)
       .subscribe(
         () => this.onSuccess(),
         (error) => this.onError(error)
       );
+
+    this.walletService.loadBalances();
```
Senyoret1: This should not go here, because it updates the balance before the transaction is sent.

src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
```diff
@@ -39,8 +42,7 @@ export class SendVerifyComponent {
 
   private onSuccess() {
     this.sendButton.setSuccess();
```
Senyoret1: In addition to this, it is necessary to add `this.sendButton.setDisabled();`

src/app/app.module.ts
alexandermushinskiy: Fixed.

src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
alexandermushinskiy: Fixed.

alexandermushinskiy:
Review comments are fixed.

src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
alexandermushinskiy: Fixed but we don't understand why it should be added. The disabled state was set before injecting transaction and not changed. 'setSuccess()' is not changed the disabled state as well.

Senyoret1: It is necessary because the disabled state is set before injecting the transaction to `backButton`, not to `sendButton`. Without the change, after sending the transaction the result is this:

enabled

In that state the user could press the button again. With the change the result is this:

disabled

PR #185, opened 2018-05-29
#172: Prevent the user from adding more than one wallet with the same seed

### LINKED ISSUE #172: Prevent the user from adding more than one wallet with the same seed
If the user tries to create or load a wallet with a seed already used, the process of creating the wallet must be stopped and an error message must be displayed.

alexandermushinskiy:
Fixes #172 

Changes:
- `walletService.create()`: add check if user creates a wallet with already existing seed;
- `walletService.create()` returns `promise<void>`, instead of `void` to have simple ability of error handling;
- update unit tests;
- update e2e tests;

### DIFF
--- a/src/app/components/pages/onboarding/onboarding-create-wallet/onboarding-create-wallet.component.ts
+++ b/src/app/components/pages/onboarding/onboarding-create-wallet/onboarding-create-wallet.component.ts
@@ -93,13 +93,13 @@ export class OnboardingCreateWalletComponent implements OnInit {
   }
 
   createWallet() {
-    this.walletService.create(this.form.value.label, this.form.value.seed);
-    this.showSafe();
+    this.walletService.create(this.form.value.label, this.form.value.seed)
+      .then(() => this.showSafe());
   }
 
   loadWallet() {
-    this.walletService.create(this.form.value.label, this.form.value.seed);
-    this.skip();
+    this.walletService.create(this.form.value.label, this.form.value.seed)
+      .then(() => this.skip());
   }
 
   skip() {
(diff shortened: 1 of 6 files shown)


alexandermushinskiy:
Tests and conflicts are fixed.

src/app/components/pages/onboarding/onboarding-create-wallet/onboarding-create-wallet.component.ts
```diff
@@ -89,8 +89,8 @@ export class OnboardingCreateWalletComponent implements OnInit {
   }
 
   loadWallet() {
-    this.walletService.create(this.form.value.label, this.form.value.seed);
-    this.skip();
+    this.walletService.create(this.form.value.label, this.form.value.seed)
+      .then(() => this.skip());
```
Senyoret1: When there is an error, it is not shown to the user.

src/app/services/wallet.service.ts
```diff
@@ -59,15 +59,27 @@ export class WalletService {
...
+
+      const wallet = {
+        label: label,
+        seed: seed,
+        addresses: [this.cipherProvider.generateAddress(this.ascii_to_hexa(seed))]
+      };
+
+      this.all.first().subscribe((wallets: Wallet[]) => {
+        if (wallets.some((w: Wallet) => w.addresses[0].address === wallet.addresses[0].address)) {
+          throw new Error('trying to add an existing wallet!');
```
Senyoret1: Please use the same error message as the desktop wallet: "A wallet already exists with this seed".

alexandermushinskiy:
Review comments are fixed.

PR #186, opened 2018-05-29
#173: Use TransactionInfoComponent in the history modal window

### LINKED ISSUE #173: Use TransactionInfoComponent in the history modal window
The content on the modal window that is opened when the user selects a transaction in the history page should be an instance of TransactionInfoComponent, just like in the desktop wallet.

alexandermushinskiy:
Fixes #173 

Changes:
- Use TransactionInfoComponent in the history modal window instead of inline html.


alexandermushinskiy:
Fixed:  Tx id is added.

PR #194, opened 2018-05-30
#187: Add e2e test(s) for "Prevent the user from adding more than one wallet with the same seed

### LINKED ISSUE #187: Add e2e test(s) for "Prevent the user from adding more than one wallet with the same seed"
Should be implemented after the #172 will be merged.

alexandermushinskiy:
Fixes #187 

Changes:
- add e2e tests

### DIFF
--- a/e2e/wallets.e2e-spec.ts
+++ b/e2e/wallets.e2e-spec.ts
@@ -56,7 +56,7 @@ describe('Wallets', () => {
     expect<any>(page.loadWalletCheckValidationLabel()).toEqual(false);
   });
 
-  it('should load wallet with already used seed', () => {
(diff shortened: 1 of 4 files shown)


e2e/wallets.e2e-spec.ts
```diff
@@ -52,6 +56,10 @@ describe('Wallets', () => {
     expect<any>(page.loadWalletCheckValidationLabel()).toEqual(false);
   });
 
+  it('should load wallet with already used seed', () => {
```
Senyoret1: The "not" is missing, it should say "should not load..."

alexandermushinskiy: Fixed.

Issue #188, opened 2018-05-30
Encode the transaction before displaying the preview


Senyoret1:
Currently, when sending a transaction, the transaction is oncoded and sent to the server when the user presses the "Send" button on the preview page. However, the transaction should be encoded (but not sended) right after the user presses the "Verify" button in the send coins form, just before displaying the preview page, in order to capture any error returned by `CipherProvider` and let the user correct it directly in the form.

alexandermushinskiy:
The issue is implemented.

PR #199, opened 2018-05-30
#190: If the transaction fails, the preview is left in an invalid state

### LINKED ISSUE #190: If the transaction fails, the preview is left in an invalid state
On the transaction preview page, if the "Send" button is pressed and the operation fails, the "Back" button remains deactivated and the amount of coins in the outputs is left multiplied by 1000000. The "Back" button must be reactivated and the amount of coins shown must not be modified.

AndreiIarovoi:
Fixes #190 

Changes:
- setEnabled() method has been added to the ButtonComponent
- OnDestroy hook has been implemented by the SendVerifyComponent component to disable toaster notifications

Comment: The problem with multiplied coins value will be fixed in scope of the issue #188

### DIFF
--- a/src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
+++ b/src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
@@ -4,6 +4,7 @@ import { MatSnackBar, MatSnackBarConfig } from '@angular/material';
 import { PreviewTransaction } from '../../../../app.datatypes';
 import { WalletService } from '../../../../services/wallet.service';
 import { ButtonComponent } from '../../../layout/button/button.component';
+import { parseResponseMessage } from './../../../../utils/errors';
 
 @Component({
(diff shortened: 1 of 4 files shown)


src/app/components/pages/send-skycoin/send-verify/send-verify.component.ts
```diff
@@ -47,9 +52,11 @@ export class SendVerifyComponent {
   }
 
   private onError(error) {
+    const errorMessage = parseResponseMessage(error['_body']);
```
Senyoret1: `error['_body']` is undefined, so an error occurs when reaching this line.

To test the errors, please modify the last line of `WalletService.injectTransaction` with `return Observable.throw(new Error("Error"));`

AndreiIarovoi: Fixed

PR #200, opened 2018-05-30
#188: Encode the transaction before displaying the preview

### LINKED ISSUE #188: Encode the transaction before displaying the preview

alexandermushinskiy:
Fixes #188 

Changes:
- rename method `ascii_to_hexa` to `convertAsciiToHexa` to be consistent in naming;
- fix the bug from the issue #190 (multiplied coins);
- move encoding of transaction by cipher provider from send-verify page to send-form page.

### DIFF
--- a/src/app/app.datatypes.ts
+++ b/src/app/app.datatypes.ts
@@ -38,6 +38,7 @@ export class NormalTransaction extends Transaction {
 export class PreviewTransaction extends Transaction {
   from: string;
   to: string;
+  row: string;
 }
 
 export interface Output {
--- a/src/app/app.datatypes.ts
+++ b/src/app/app.datatypes.ts
@@ -38,7 +38,7 @@ export class NormalTransaction extends Transaction {
 export class PreviewTransaction extends Transaction {
   from: string;
   to: string;
-  row: string;
+  encoded: string;
 }
 
 export interface Output {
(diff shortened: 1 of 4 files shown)


src/app/app.datatypes.ts
```diff
@@ -38,6 +38,7 @@ export class NormalTransaction extends Transaction {
 export class PreviewTransaction extends Transaction {
   from: string;
   to: string;
+  row: string;
```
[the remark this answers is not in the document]
alexandermushinskiy: Renamed to "encoded".

src/app/services/wallet.service.ts
```diff
@@ -187,18 +187,14 @@ export class WalletService {
           inputs: txInputs,
           outputs: txOutputs,
           hoursSent: hoursToSend,
-          hoursBurned: totalHours - calculatedHours
+          hoursBurned: totalHours - calculatedHours,
+          row: this.generateRawTransaction(txInputs, txOutputs)
```
Senyoret1: The name must be changed here too, or it will not be possible to send transactions, because `encoded` will always be `undefined`.

alexandermushinskiy: Sorry for the inattention, fixed.

PR #206, opened 2018-05-31
#205: Prod build error

### LINKED ISSUE #205: Prod build error
When using `npm run build` there are some compilation errors that must be fixed. In addition, `.travis.yml` must be modified so that Travis performs a production build as part of the tests it normally performs.

alexandermushinskiy:
Fixes #205 

Changes:
- update `transaction-info` component: show balance only if it's not preview;
- update `.travis.yaml`: add script to perform a production build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ addons:
 
 script:
   - make check
+  - make build
 
 notifications:
   # https://github.com/kvld/travisci-telegram TravisCI Telegram Bot integration
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 .DEFAULT_GOAL := help
 .PHONY: lint check help
 
+build:
+	npm run build
+
 lint: ## runs lint
 	npm run lint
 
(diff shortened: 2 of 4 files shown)


alexandermushinskiy:
@Senyoret1, yes, you are right, the error because Transaction class didn't have property `balance`.
The `balance` is not calculated when user wants to send coins: this field is not used neither `createTransaction()` nor `SendFormComponent`, so in preview it is undefined, isn't it?

Senyoret1:
Indeed, it is a bit difficult to find and it looks like it should be undefined, but it is not, `balance` is assigned here:
https://github.com/skycoin/skycoin-web/blob/469f0406d487afef6775fcdb2e2958ad20caa67c/src/app/components/pages/send-skycoin/send-skycoin.component.ts#L25-L33

alexandermushinskiy:
I need to fix my attentiveness urgently. It has been fixed.

Senyoret1:
No problem, that happens sometimes :smile:

PR #226, opened 2018-06-01
Fix balance update problems

### LINKED ISSUE #225: Additional problems while updating the balance
When there is an error in the connection to the server to update the balance, the procedure of updating the balance periodically is completelly stoped, until the user refreshes the browser.

Senyoret1:
Fixes #225

Changes:
- If there is an error when connecting to the server, the balance will be updated again after 20 seconds (the timer is restarted as if there were pending transactions).
- The number of minutes that has passed since the last balance update is rounded down.


AndreiIarovoi:
@Senyoret1, we don't have rights (we are not project collaborators) to merge the PR.

PR #244, opened 2018-06-04
#191: Improvements for ConfirmationComponent

### LINKED ISSUE #191: Improvements for ConfirmationComponent
The text content of `ConfirmationComponent` should be centered.

Also, when using `ConfirmationComponent` to confirm if the user really wants to delete a wallet, it should show a `<mat-checkbox>` (like the one used in `OnboardingSafeguardComponent`) that the user must select before being able to confirm the deletion of the wallet, as an extra security measure (taking into account that many users do not read the warnings).

AndreiIarovoi:
Fixes #191 

Changes:
- Added Confirmation to confirm that the user really wants to delete a wallet
- Unit tests has been adjusted
- Minor refactoring of involved components has been done


AndreiIarovoi:
Fixed

src/app/components/pages/wallets/address-detail/wallet-detail.component.ts
```diff
@@ -94,7 +96,11 @@ export class WalletDetailComponent implements OnInit {
   deleteWallet() {
     const dialogRef = this.dialog.open(ConfirmationComponent, {
       width: '500px',
-      data: { text: `${this.deleteConfirmation1} "${this.wallet.label}" ${this.deleteConfirmation2}` }
+      data: {
+        text: `${this.deleteConfirmation1} "${this.wallet.label}" ${this.deleteConfirmation2}`,
+        isCheckboxDisplayed: true,
```
Senyoret1: Please change the name of the property from `isCheckboxDisplayed` to `displayCheckbox`.

AndreiIarovoi: Fixed

PR #245, opened 2018-06-04
#240: Indicate if there was a problem updating the balance 

### LINKED ISSUE #240: Indicate if there was a problem updating the balance
If there is an error while updating the balance, place an blinking icon in the header, like this:

problem

The icon should show a tooltip with the following text: "Connection error, retrying"

The icon in the image is https://material.io/tools/icons/?icon=warning&style=baseline

AndreiIarovoi:
Fixes #240 

Changes:
- Blinking icon in the header added in case any issue with updating the balance
-  Tooltip with the "Connection error, retrying" text has been added on the warning item
- Unit tests have been adjusted

### DIFF
--- a/src/app/components/layout/header/top-bar/top-bar.component.scss
+++ b/src/app/components/layout/header/top-bar/top-bar.component.scss
@@ -9,7 +9,7 @@
 .buttons-left {
   display: inline-block;
   padding: 0 10px;
-  width: 220px;
+  width: 205px;
 }
 
 .last-updated-time {
@@ -31,7 +31,7 @@
 .buttons-right {
   display: inline-block;
   text-align: right;
-  width: 200px;
+  width: 205px;
 
   button {
     display: inline-block;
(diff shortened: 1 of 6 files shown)


src/app/services/wallet.service.ts
```diff
@@ -439,7 +443,7 @@ export class WalletService {
   }
 
   private resetBalancesTimerOptions(hasPendingTxs: boolean) {
-    this.intervalTime = (hasPendingTxs ? 20 : 60) * 1000;
+    this.intervalTime = (hasPendingTxs ? 20 : 300) * 1000;
```
Senyoret1: Do not change this or the label with how many minutes have passed since the last update will stop updating the time correctly.

src/app/components/layout/header/top-bar/top-bar.component.scss
```diff
@@ -9,7 +9,7 @@
 .buttons-left {
   display: inline-block;
   padding: 0 10px;
-  width: 200px;
+  width: 220px;
```
Senyoret1: If you change this, you also have to change the width of `.buttons-right`. Otherwise the title is stops being centered correctly.

AndreiIarovoi:
Fixed

PR #258, opened 2018-06-06
Sky#247

### LINKED ISSUE #247: Use ngx-translate in the places where it is missing.
At this time most of the texts use ngx-translate. However, it is necessary to perform a thorough revision of the code to detect the parts in which it was not used. For example, it is missing here:
https://github.com/skycoin/skycoin-web/blob/efd0d57a34bf9db1dcbd7d3d75240595dbfc48b2/src/app/components/pages/send-skycoin/send-form/send-form.component.html#L10
And here:
https://github.com/skycoin/skycoin-web/blob/efd0d57a34bf9db1dcbd7d3d75240595dbfc48b2/src/app/components/pages/send-skycoin/send-verify/transaction-info/transaction-info.component.html#L23
It is important to make an exhaustive revision of all the .html and .ts files, including the pipes and the services.

AndreiIarovoi:
Fixes #247

Changes:
- ngx-translate component has been used for all missing pages components
- ngx-translate component has been used for all layouts, pipes and services
- ConfirmationComponent has been changed to receive all its parameters through data parameter
- ConfirmationData interface has been added to the "app.datatypes.ts"
- SkyPipe has been removed since it is not used any more
- Unit tests have been adjusted
- Minor refactoring of the involved components


src/app/services/api.service.ts
```diff
@@ -83,4 +85,11 @@ export class ApiService {
   private getCsrf() {
     return this.get('csrf').map(response => response.csrf_token);
   }
+
+  private getErrorMessage(error: any): Observable<any> {
+    return error
+      ? Observable.throw(error)
+      : this.translate.get('service.api.server-error')
+          .flatMap(message => Observable.throw(message));
```
Senyoret1: For this to work properly, it must be modified to `.flatMap(message => Observable.throw(new Error(message)));`

AndreiIarovoi: Fixed

PR #268, opened 2018-06-08
#253: Additional unit tests

### LINKED ISSUE #253: Additional unit tests
- In `cipher.provider.spec.ts` should be tested that when creating a transaction with an invalid destination address an error must be thrown.

- The tests in `cipher.provider.spec.ts` should be done again, but calling the functions from `WalletService`. Probably the best thing is to create a new unit tests file for `WalletService`, in which a real instance of `CipherProvider` is also used, not a simulated one.

- `WalletService` should test the `delete` function.

AndreiIarovoi:
Fixes #253 Additional unit tests 

Changes:
- "Creating a transaction with an invalid destination address an error must be thrown" unit test has been added to the cipher.provider.spec.ts 
- "Delete wallet" unit test has been added to the wallet.service.spec.ts.
- Additional "service.cipher.spec.ts" unit test file has been added to the for the walletе service to perform tests of cipher.provider.spec.ts which uses a real instance of CipherProvider.


src/app/services/wallet.service.cipher.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+
+      const outputs: Output[] = [
+        createOutput(addresses[0].address, amount, 1),
+      ];
+
+      spyApiService.getOutputs.and.returnValue(Observable.of(outputs));
+
+      walletService.createTransaction(wallet, wrongDestinationAddress, amount)
+        .subscribe(
+          () => {},
```
Senyoret1: The test must fail if this event is executed.

src/app/services/wallet.service.cipher.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+      ];
+
+      spyApiService.getOutputs.and.returnValue(Observable.of(outputs));
+
+      walletService.createTransaction(wallet, destinationAddress, amount)
+        .subscribe(
+          (result: any) => {
+          expect(result.inputs).toEqual(expectedTxInputs);
+          expect(result.outputs).toEqual(expectedTxOutputs);
+        });
```
Senyoret1: The test must fail if the error event is executed.

PR #269, opened 2018-06-08
#262: Improvements for the e2e tests of the onboarding page

### LINKED ISSUE #262: Improvements for the e2e tests of the onboarding page
- A test should be added to make sure that the create new wallet button is not activated if the seeds do not match.
- `createWallet` and `createExistingWallet` repeat the same code to fill out the form. The code should be in a separate function, to avoid repeating it. This also happens with `loadWallet` and `loadExistingWallet`.
- After `should hide accepted safeguard` a test should be added to verify that the created wallet has the correct address.
- The buttons for creating seeds with 12 and 24 words must be tested.
- A test should be added to load a wallet and check if the loaded wallet has the correct address.
- A test should be added to make sure that the disclaimer modal window can not be closed by pressing the X button or by clicking on the dark background behind the modal window.
- A test should be added to make sure that the safeguard modal window can be closed by pressing the X button and by clicking on the dark background behind the modal window.

alexandermushinskiy:
Fixes #262 

Changes:
- added additional e2e tests of the onboarding page;
- removed code duplication in `onboarding-create.po.ts`;


alexandermushinskiy:
But the test `should load wallet` just checks if `Create` button is enabled, it's not load wallet and save it to the local storage.

Senyoret1:
Yes, but it also fills the form, so it could be a starting point for testing if the walllet is created with the expected address, just like `should create wallet` is used as a starting point for creating the wallet. However, in any case it is only a suggestion.

alexandermushinskiy:
Yes, it fills the form and check if button is enabled, after that `should create wallet` test is called, form resets with values and `Create` button is clicked (means wallet is created). The problem is that on `onboarding` page we can only create or load wallet, we cannot do both actions due to user will be redirected to `wallets` page after wallet is created or loaded.

Senyoret1:
Well, if that test gives a lot of trouble it can be left out for now. That specific case may be reconsidered later, in a review that I have planned to do in a few days, so no big deal.

alexandermushinskiy:
As a proposition we can create separate file `onboarding-load.e2e-spec.ts` to have whole loading flow. What do you think?

Senyoret1:
Ok, no problem, but in a separate pr please. I'm going to merge this, since the changes made here work correctly.

alexandermushinskiy:
Ok, let push my changes regarding `it should simply be tested if the disclamer shows the X button`.

Senyoret1:
Wow, I merged it before your comment. Please, add the changes to a new PR, it can be a PR only with that change, and sorry for the inconvenience.

alexandermushinskiy:
The PL is ready. Also we added the test `load a wallet and check if the loaded wallet has the correct address`.

Senyoret1:
Great, thank you very much.

PR #270, opened 2018-06-08
#162: Unlock your wallet in an asynchronous procedure

### LINKED ISSUE #162: Unlock your wallet in an asynchronous procedure
After entering the seed in the unlock wallet modal window and pressing the "Unlock" button, the browser freezes, which is not good for the UX. To avoid this, the wallet must be unlock in an asynchronous process and the "Unlock" button should show the loading animation.

alexandermushinskiy:
Fixes #162

Changes:
- added using web worker to unlock wallet;
- updated unit and e2e tests;

### DIFF
--- a/src/app/services/wallet.service.ts
+++ b/src/app/services/wallet.service.ts
@@ -6,6 +6,8 @@ import 'rxjs/add/operator/filter';
 import 'rxjs/add/operator/first';
 import 'rxjs/add/operator/mergeMap';
 import 'rxjs/add/observable/zip';
+import 'rxjs/add/observable/range';
+import 'rxjs/add/observable/from';
 import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 import { Observable } from 'rxjs/Observable';
 import { Subject } from 'rxjs/Subject';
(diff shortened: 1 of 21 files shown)


src/app/services/wallet.service.ts
```diff
@@ -177,22 +178,22 @@ export class WalletService {
...
-        address.public_key = fullAddress.public_key;
-        currentSeed = fullAddress.next_seed;
-      });
+    return new Promise<void>((resolve, reject) => {
+      const currentSeed = this.convertAsciiToHexa(seed);
 
-      wallet.seed = seed;
-      this.updateWallet(wallet);
-      return resolve();
+      WebWorkersHelper.ExcecuteWorker('/assets/scripts/workers/generate-addresses.worker.js', { seed: currentSeed, addresses: wallet.addresses })
```
Senyoret1: I think that `WebWorkersHelper.ExcecuteWorker` should be in `CipherProvider.generateAddress` and not here. I think so because with this change we now have two different codes to generate addresses: the one in `CipherProvider.generateAddress` and the one in `generate-addresses.worker.js`. To not have duplicate code, it is necessary to choose one of the two, but `generate-addresses.worker.js` has some problems, like the fact that it has no tests (while `CipherProvider` has unit tests and the test suite adapted from Go) and it is in Javascript (which is inconvenient due to the lack of consistency with the rest of the code, the lack of compilation errors, etc.).

Bearing that in mind, and the fact that it will eventually be necessary to also adapt `CipherProvider.prepareTransaction`, I think that keping the code in the `.worker.js` files to the bare minimun (only to call the functions from `main.js`) and keep all the logic inside `CipherProvider` is much more convenient.

alexandermushinskiy: Would it be OK to have 2 separate web workers: 1- generate single address; 2 - generate the list of addresses?

Senyoret1: I'm not sure why that would be needed. `CipherProvider` does not have a function to create a list of addresses, it only creates one address at a time. `Cipher` from `main.js` have a function called `GenerateAddresses` but it just creates one address.

If you want to create a new function that returns multiple addresses, you can simply use TypeScript to call several times the web worker that creates one address, that would not negatively affect the UX, so it would not be necessary to create a web worker specifically for that. The idea is to write as little JavaScript code as possible.

PR #283, opened 2018-06-12
#281: Use only one convertAsciiToHexa function

### LINKED ISSUE #281: Use only one convertAsciiToHexa function
Currently there are different `convertAsciiToHexa` functions (probably some with a different name) in the code. It is necessary to delete all the repeated versions of the function (both in the application code and in the test code) and use only one, which must be inside `/src/app/utils/`.

It is important to use the version in `cipher.provider.lib.spec.ts`, because, unlike the others, it is modified to correctly process the null characters.

AndreiIarovoi:
Fixes #281 

Changes:
- Repeated convertAsciiToHexa function has been moved to the "\src\app\utils\converters.ts" file


AndreiIarovoi:
Fixed

PR #285, opened 2018-06-12
#282: show ellipsis if wallet label is too long on wallet page

### LINKED ISSUE #282: Wallet label is splitted up on two rows if length is too long
At the pic below the wallet label is `skycoin-web-e2e-test-load-wallet-seed`
image

@Senyoret1, should we fix it somehow?
One of possible solution is to show ellipsis at the end and show tooltip with a full string.

alexandermushinskiy:
Fixes #282 

Changes:
- in css to show ellipsis if wallet label is too long on wallet page.


src/app/components/pages/wallets/wallets.component.html
```diff
@@ -10,7 +10,7 @@
     <div class="-wallets">
       <ng-container *ngFor="let wallet of wallets">
         <div class="-wallet" (click)="toggleWallet(wallet)">
-          <div class="-width-250 -label">{{ wallet.label }}</div>
+          <div class="-width-250 -label" [attr.title]="wallet.label.length > 36 ? wallet.label : ''">{{ wallet.label }}</div>
```
Senyoret1: Please remove the 36 characters condition. If the name has many wide letters (like "M"), 27 characters could be enough to show the ellipsis, but if it has many narrow letters (like "i"), 47 characters fit without problems. As it is so difficult to know how many characters fit in the space, it seems that it is best to show the tooltip always, instead of risking to have an inconsistent UI.

alexandermushinskiy: Fixed.

PR #286, opened 2018-06-12
#280: It should be possible to send the forms by pressing enter

### LINKED ISSUE #280: It should be possible to send the forms by pressing enter
Please, use https://github.com/skycoin/skycoin/pull/1379 as reference. Note that in the reference PR some important details are discussed, such as that it should not be possible to send the forms by pressing enter if the last field is a textarea (there is even a point commenting the last field of the send coins form is a textarea, something that is no longer true) and that it is necessary to be careful so that it is not possible to send the form several times when pressing enter quickly.

alexandermushinskiy:
Fixes #280 

Changes:
- added ability to submit form by pressing enter in following components: ChangeNameComponent, SendFormComponent, OnboardingEncryptWalletComponent.


alexandermushinskiy:
The issue is implemented.
Do we need OnboardingEncryptWalletComponent component? User can only be navigated there by specifying the appropriate url manually.

Senyoret1:
There is a compilation error and the branch must be updated.

`OnboardingEncryptWalletComponent` will probably be used in the future, so I would prefer not to delete it for now.

alexandermushinskiy:
Fixed.

PR #298, opened 2018-06-14
Reuse the web worker

### LINKED ISSUE #297: Reuse the web worker
Currently, a web worker is created and destroyed every time an address has to be created. The web worker takes a long time to initialize because it must load the `Cipher` library, which is very large, so the current process is extremely inefficient. To improve efficiency, a single web worker must be created and reused whenever necessary.

More details about the problem in https://github.com/skycoin/skycoin-web/issues/162#issuecomment-396798374

Senyoret1:
Fixes #297 

Changes:
- `generate-addresses.worker.js` was converted into `cipher-web-worker.js`. The idea is to have a single web worker for all the functions in `Cipher`, so it can be reused efficiently. The web worker is still limited to calling functions from `Cipher`, with as little code as possible. The difference is that it now receives a parameter indicating which function it should call. With this change, the web worker is not only able to call `Cipher.GenerateAddresses`, but also `Cipher.PrepareTransaction`.

- `web-workers-helper.ts` was converted into `cipher-web-worker-helper.ts`. Now it is simply responsible for maintaining and reusing an instance of `cipher-web-worker.js`. The instance is created right after the wallet is started, so its functions start running quickly.

- To ensure that everything works correctly with more than one operation at a time, `cipher-web-worker-helper.ts` sends an ID as a parameter to `cipher-web-worker.js`, and `cipher-web-worker.js` must return that same ID when the operation finishes.

With these changes, the calls to the web worker are almost as fast as before.


src/assets/scripts/cipher-web-worker.js
```diff
@@ -0,0 +1,17 @@
+importScripts('main.js');
+
+onmessage = function(e) {
+  try {
+    if (e.data.operation == 0) {
```
alexandermushinskiy: This would be much better represented as a `switch-case-default` as it then makes it immediately obvious that the choice of action is occurring based on the value of "value" and not some arbitrary test.

src/app/utils/cipher-web-worker-helper.ts
```diff
@@ -0,0 +1,48 @@
...
+  private static readonly activeWorks: Map<number, Subject<any>> = new Map<number, Subject<any>>();
+  private static initialized = false;
+
+  static ExcecuteWorker(operation: CipherWebWorkerOperations, data: any): Observable<any> {
+
+    if (!CipherWebWorkerHelper.initialized) {
+      CipherWebWorkerHelper.initialized = true;
+
+      CipherWebWorkerHelper.worker.addEventListener('message', (e: MessageEvent) => {
+        if (isString(e.data.result) && (e.data.result as string).startsWith(CipherWebWorkerHelper.errPrefix)) {
```
alexandermushinskiy: As a proposition, we can make the method `ExcecuteWorker` less complicated to move the event listener body to the separate method. In that case it will look like:
`CipherWebWorkerHelper.worker.addEventListener('message', (e: MessageEvent) => eventListener(e.data));`

PR #307, opened 2018-06-14
#304: Create transactions with the web worker

### LINKED ISSUE #304: Create transactions with the web worker
Create the transactions in `CipherProvider.prepareTransaction` using the web worker. The code should be something like:

```
CipherWebWorkerHelper.ExcecuteWorker(CipherWebWorkerOperations.PrepareTransaction, {inputs: JSON.stringify(inputs), outputs: JSON.stringify(outputs)})
```

Also, with this change `CipherProvider` will no longer use `Cipher` directly, so `declare var Cipher;` should be deleted.

After making the change, verify if the UI correctly shows the spinner while the transaction is being created.

alexandermushinskiy:
Fixes #304 

Changes:
- updated transaction creation by using web worker

### DIFF
--- a/src/app/services/cipher.provider.spec.ts
+++ b/src/app/services/cipher.provider.spec.ts
@@ -39,23 +39,13 @@ describe('CipherProvider', () => {
       createTransactionOutput('2e1erPpaxNVC37PkEv3n8PESNw2DNr5aJNy', 20000)
     ];
 
-    const actualLength = cipherProvider.prepareTransaction(inputs, outputs).length;
-    const expectedLength = 440;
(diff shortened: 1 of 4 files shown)


alexandermushinskiy:
`should be rejected for an invalid destination address` is a possible behaviour of Cipher which was called directly in provider. Now we have a chain: provider-web worker-Cipher and it looks as not truly unit-test of the provider. That's why we decided to remove that unit-test.

Other tests, except `should be created`, do not work properly. They are marked as passed but only because there is no error during test body flow, the test does not wait when `subscribe` is called it just finishes. 

Regarding the testing of errors, we added tests just to check if error occurs in web worker and it will be thrown in provider.

PR #318, opened 2018-06-15
Fix the icons of the wallet options buttons

### LINKED ISSUE #317: Fix the icons of the wallet options buttons
- The background of the `minus-red.png` file must be removed.
- Create the bitmaps for the Delete Wallet button.

Senyoret1:
Fixes #317 

Changes:
- Remove the background from `minus-red.png`.
- Replace the bitmap of the Delete Wallet button with a new one (previously it used the same bitmap as the Hide Empty button).

The bitmaps were saved with web optimizations. The button bar now looks like this:
buttons
The new Delete Wallet icon is not inside a circle, like the ones on the left, because otherwise the Edit Wallet icon would look out of place.


alexandermushinskiy:
No problem, it's done.

Senyoret1:
Thanks

PR #321, opened 2018-06-18
#271: Improvements for the e2e tests of the wallets page

### LINKED ISSUE #271: Improvements for the e2e tests of the wallets page
- Many functions repeat almost the same code to fill in the create/load wallet form. There should be a function to fill in the form, in such a way that it is only necessary to send the content of each field as a parameter.

- `should create wallet` and `should load wallet` do not check if the wallet was created, but if the button to create it was enabled. It is necessary to check if the wallet was created and if it has the correct address.

- In `should show wallet QR modal`, check if the modal window shows the correct address.

- In `should add address to wallet` and `should unlock wallet component on add new address for locked wallet`, check if the correct address was added.

- Test the "Delete Wallet" button.

alexandermushinskiy:
Fixes #271

Changes:
- added e2e tests for wallets page


e2e/wallets.po.ts
```diff
@@ -24,119 +24,62 @@ export class WalletsPage {
...
-            });
-          });
-        });
-      });
-    });
+    this.fillWalletForm('Test wallet', 'test test2', 'skycoin-web-e2e-test-seed');
+    return btnCreate.isEnabled();
   }
 
   loadWalletCheckValidationSeed() {
```
Senyoret1: Now this function does not check the condition correctly. Before the changes it put a label and left the seed empty, to check if the seed was being validated correctly. Now it leaves both the label and the seed empty, so the validation of the seed could be failing, but the test would pass correctly because the button would be disabled because the label is empty.

e2e/wallets.po.ts
```diff
@@ -280,14 +228,95 @@ export class WalletsPage {
...
+
+    return element(by.buttonText('No')).click().then(() => {
+      return element.all(by.css('.-wallet')).count().then((count) => {
+        return count === walletCount;
+      });
+    });
+  }
+
+  deleteWallet() {
+    const walletNameToDelete = 'Test Create Name';
```
Senyoret1: This name is not used previously, so a wallet with this name will never be found, regardless of whether the delete operation was successful or not.

e2e/wallets.po.ts
```diff
@@ -280,14 +228,95 @@ export class WalletsPage {
...
+
+    return element.all(by.css('.btn-delete-wallet')).first().click().then(() => {
+      return element(by.css('.-disclaimer-check-text')).click().then(() => {
+        return element(by.buttonText('Yes')).click().then(() => {
+          return element.all(by.css('.-wallet')).count().then(() => {
```
Senyoret1: It does not seem necessary to count the number of wallets, because the number is never used.

e2e/wallets.po.ts
```diff
@@ -280,14 +228,95 @@ export class WalletsPage {
...
+            return this.isWalletExist(walletNameToDelete).then((isWalletExistAfter) => {
+              return !isWalletExistAfter;
+            });
+          });
+        });
+      });
+    });
+  }
+
+  private isWalletExist(name: string) {
```
Senyoret1: It would be better to change the name to simply `walletExist`

e2e/wallets.po.ts
alexandermushinskiy: Fixed, rolled it back.

e2e/wallets.po.ts
alexandermushinskiy: Fixed.

e2e/wallets.po.ts
alexandermushinskiy: Fixed.

PR #324, opened 2018-06-18
#308: Create CoinService

### LINKED ISSUE #308: Create CoinService
It is necessary to create a service for controlling the coins that are compatible with the wallet. In general, the service should have the following:

- An array in which all compatible coins can be consulted. The array must be filled with the objects contained in the files of the `src/app/coins` folder.
- A property with the currently selected coin. The currently selected coin must be a `BaseCoin` object. The property should be a `BehaviorSubject`, to be able to know when the coin has ben changed.
- A function for changing the current coin. The selected coin must be saved in permanent storage when calling the function, in order to be selected again when opening the wallet again.

alexandermushinskiy:
Fixes #308 

Changes:
- added coin service.

### DIFF
--- a/src/app/services/coin.service.ts
+++ b/src/app/services/coin.service.ts
@@ -1,0 +1,40 @@
+import { BehaviorSubject } from 'rxjs/BehaviorSubject';
+
+import { BaseCoin } from '../coins/basecoin';
+import { SkycoinCoin } from '../coins/skycoin.coin';
+import { TestCoin } from '../coins/test.coin';
+
+export class CoinService {
+
+  currentCoin: BehaviorSubject<BaseCoin> = new BehaviorSubject<BaseCoin>(null);
+
+  private coins: BaseCoin[] = [];
+  private readonly storageKey = 'currentCoin';
+
+  constructor() {
+    this.loadCurrentCoin();
+    this.loadAvailableCoins();
+  }
+
+  changeCoin(coin: BaseCoin) {
+    this.currentCoin.next(coin);
+
+    this.saveCoin(coin);
+  }
+
+  private loadCurrentCoin() {
+    const storedCoin: string = localStorage.getItem(this.storageKey);
+    if (storedCoin) {
+      const coin: BaseCoin = JSON.parse(storedCoin);
+      this.currentCoin.next(coin);
+    }
+  }
+
+  private saveCoin(coin) {
+    localStorage.setItem(this.storageKey, JSON.stringify(coin));
+  }
+
+  private loadAvailableCoins() {
+    this.coins.push(new SkycoinCoin(), new TestCoin());
+  }
+}
--- a/src/app/services/coin.service.ts
+++ b/src/app/services/coin.service.ts
@@ -3,7 +3,7 @@ import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 import { BaseCoin } from '../coins/basecoin';
 import { SkycoinCoin } from '../coins/skycoin.coin';
 import { TestCoin } from '../coins/test.coin';
-import { coinsId } from '../constants/coins-id.const';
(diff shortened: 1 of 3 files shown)


src/app/services/coin.service.ts
```diff
@@ -0,0 +1,40 @@
...
+  private loadCurrentCoin() {
+    const storedCoin: string = localStorage.getItem(this.storageKey);
+    if (storedCoin) {
+      const coin: BaseCoin = JSON.parse(storedCoin);
+      this.currentCoin.next(coin);
+    }
+  }
+
+  private saveCoin(coin) {
+    localStorage.setItem(this.storageKey, JSON.stringify(coin));
```
Senyoret1: Only the coin ID should be saved in persistent storage, not the entire object. An example in which this would be necessary is if the URL of the node changes: if the whole object is saved and restored, updating the URL in the object of the coin would not work because the URL would be obtained from the persistent storage, no the updated object.

src/app/services/coin.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    this.saveCoin(coin);
+  }
+
+  private loadCurrentCoin() {
+    const storedCoin: string = localStorage.getItem(this.storageKey);
+    if (storedCoin) {
+      const coin: BaseCoin = JSON.parse(storedCoin);
+      this.currentCoin.next(coin);
+    }
```
Senyoret1: If it is not possible to obtain a coin from persistent storage, `this.currentCoin` should be set to the default coin.

src/app/services/coin.service.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { BaseCoin } from '../coins/basecoin';
+import { SkycoinCoin } from '../coins/skycoin.coin';
+import { TestCoin } from '../coins/test.coin';
+
+export class CoinService {
+
+  currentCoin: BehaviorSubject<BaseCoin> = new BehaviorSubject<BaseCoin>(null);
+
+  private coins: BaseCoin[] = [];
```
Senyoret1: It should be possible to consult the contents of this array. It will be used to fill in the list of available coins.

src/app/services/coin.service.ts
alexandermushinskiy: Fixed.

src/app/services/coin.service.ts
alexandermushinskiy: Fixed.

src/app/services/coin.service.ts
alexandermushinskiy: @Senyoret1, what do you mean "to consult"? Do you mean to make it public?
This array is filled in in `loadAvailableCoins()` method.

Senyoret1: Yes, it have to be public for future use, sorry for the lack of clarity.

alexandermushinskiy: Fixed.

src/app/services/coin.service.ts
```diff
@@ -3,35 +3,37 @@ import { BehaviorSubject } from 'rxjs/BehaviorSubject';
...
+import { coinsId } from '../constants/coins-id.const';
 
 export class CoinService {
 
   currentCoin: BehaviorSubject<BaseCoin> = new BehaviorSubject<BaseCoin>(null);
+  coins: BaseCoin[] = [];
 
-  private coins: BaseCoin[] = [];
   private readonly storageKey = 'currentCoin';
+  private readonly defaultCoinId = coinsId.sky;
```
Senyoret1: The defailt coin is `defaultCoinId`, from `../constants/coins-id.const`.

alexandermushinskiy: Fixed.

PR #326, opened 2018-06-18
#314: Indicate the status of the secondary pages

### LINKED ISSUE #314: Indicate the status of the secondary pages
Currently, when there are no transactions, the History page simply shows the gray background and a slightly darker line. It would be more convenient to show a component with a message for the user like this:

msg
Additionally, the same component could be used while the information is being loaded, it would only be necessary to replace the text and the icon (with the spinner). In this way the user would have more information about what the wallet is doing, which would improve the UX.

The idea would be to use the component not only on the History page, but also on the Blockchain, Outputs and Pending Transactions pages. To be able to add it correctly in the Blockchain page, it would be necessary to wait for `blockchainService.lastBlock` and `blockchainService.coinSupply` to finish, before displaying the information on the screen.

The icon used in the image is https://material.io/tools/icons/?icon=announcement&style=baseline, its color is `#eaeaea` and have a size of about 50x50 px. The The color of the text is `#a9aaac`. The content has a margin of 50px with respect to the navigation bar (a similar margin should be used at the bottom).

alexandermushinskiy:
Fixes #314 

Changes:
- added new component `loading-content`;
- show `loading-content` component when data is loading or no data.

### DIFF
--- a/src/app/components/layout/loading-content/loading-content.component.html
+++ b/src/app/components/layout/loading-content/loading-content.component.html
@@ -1,0 +1,10 @@
+<div class="content">
+  <div *ngIf="isLoading">
+    <mat-spinner [color]="spinnerStyle" class="in-button big"></mat-spinner>
+    <p>Loading...</p>
+  </div>
+  <div *ngIf="!isLoading">
+    <i class="material-icons">announcement</i>
+    <p>There are still no transactions to show</p>
+  </div>
+</div>
--- a/src/app/components/layout/loading-content/loading-content.component.html
+++ b/src/app/components/layout/loading-content/loading-content.component.html
@@ -1,10 +1,10 @@
 <div class="content">
   <div *ngIf="isLoading">
     <mat-spinner [color]="spinnerStyle" class="in-button big"></mat-spinner>
-    <p>Loading...</p>
+    <p>{{ 'common.loading' | translate }}</p>
   </div>
   <div *ngIf="!isLoading">
     <i class="material-icons">announcement</i>
-    <p>There are still no transactions to show</p>
+    <p>{{ noDataText | translate }}</p>
   </div>
 </div>
--- a/src/app/components/layout/loading-content/loading-content.component.ts
+++ b/src/app/components/layout/loading-content/loading-content.component.ts
@@ -1,0 +1,10 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'app-loading-content',
+  templateUrl: './loading-content.component.html',
+  styleUrls: ['./loading-content.component.scss']
+})
+export class LoadingContentComponent {
+  @Input() isLoading = false;
+}
--- a/src/app/components/layout/loading-content/loading-content.component.ts
+++ b/src/app/components/layout/loading-content/loading-content.component.ts
@@ -6,5 +6,6 @@ import { Component, Input } from '@angular/core';
   styleUrls: ['./loading-content.component.scss']
 })
 export class LoadingContentComponent {
-  @Input() isLoading = false;
+  @Input() isLoading = true;
+  @Input() noDataText = 'tx.none';
 }
--- a/src/app/components/pages/settings/blockchain/blockchain.component.ts
+++ b/src/app/components/pages/settings/blockchain/blockchain.component.ts
@@ -1,6 +1,7 @@
 import { Component, OnInit } from '@angular/core';
 
 import { BlockchainService } from '../../../../services/blockchain.service';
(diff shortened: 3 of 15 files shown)


src/app/components/pages/settings/blockchain/blockchain.component.ts
```diff
@@ -9,13 +10,24 @@ import { BlockchainService } from '../../../../services/blockchain.service';
 export class BlockchainComponent implements OnInit {
   block: any;
   coinSupply: any;
+  isLoading = false;
```
Senyoret1: This is not really necessary, checking `!(block && coinSupply)` is enough, similar to how it is done in `HistoryComponent`.

src/app/components/pages/settings/outputs/outputs.component.ts
```diff
@@ -13,6 +13,7 @@ import { QrCodeComponent } from '../../../layout/qr-code/qr-code.component';
 })
 export class OutputsComponent implements OnInit {
 
+  isLoading = false;
```
Senyoret1: This is not really necessary, checking `!wallets` is enough, similar to how it is done in `HistoryComponent`.

src/app/components/pages/settings/pending-transactions/pending-transactions.component.ts
```diff
@@ -37,13 +38,16 @@ export class PendingTransactionsComponent implements OnInit, OnDestroy {
   }
 
   private loadTransactions(value: number) {
+    this.isLoading = true;
+
     const showAllTransactions = value === DoubleButtonActive.RightButton;
     this.walletService.getAllPendingTransactions()
```
Senyoret1: Please, delay the execution of this 32ms. The reason is that if the server responds very quickly, pressing the "Mine" and "All" buttons will not show the loading animation, so the user might think that the wallet did nothing and that it is working incorrectly.

src/app/components/layout/loading-content/loading-content.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'app-loading-content',
+  templateUrl: './loading-content.component.html',
+  styleUrls: ['./loading-content.component.scss']
+})
+export class LoadingContentComponent {
+  @Input() isLoading = false;
```
Senyoret1: This should be `false` by default, that way it would not be necessary to assign a value to this property on pages that only use this control to display the load animation (like BlockchainComponent).

src/app/components/layout/loading-content/loading-content.component.html
```diff
@@ -0,0 +1,10 @@
+<div class="content">
+  <div *ngIf="isLoading">
+    <mat-spinner [color]="spinnerStyle" class="in-button big"></mat-spinner>
+    <p>Loading...</p>
+  </div>
+  <div *ngIf="!isLoading">
+    <i class="material-icons">announcement</i>
+    <p>There are still no transactions to show</p>
```
Senyoret1: The text should be obtained from ngx-translate.

src/app/components/layout/loading-content/loading-content.component.html
```diff
@@ -0,0 +1,10 @@
+<div class="content">
+  <div *ngIf="isLoading">
+    <mat-spinner [color]="spinnerStyle" class="in-button big"></mat-spinner>
+    <p>Loading...</p>
```
Senyoret1: The text should be obtained from ngx-translate.

src/app/components/pages/settings/blockchain/blockchain.component.ts
alexandermushinskiy: Fixed

src/app/components/pages/settings/outputs/outputs.component.ts
alexandermushinskiy: Fixed.

src/app/components/layout/loading-content/loading-content.component.html
alexandermushinskiy: Fixed.

src/app/components/layout/loading-content/loading-content.component.html
alexandermushinskiy: Fixed.

src/app/components/layout/loading-content/loading-content.component.ts
alexandermushinskiy: Fixed.

PR #362, opened 2018-06-25
#351: Problems obtaining the balance after changing the coin

### LINKED ISSUE #351: Problems obtaining the balance after changing the coin
This issue has two parts. The first is a problem that happens when changing the coin, the steps to reproduce it are: open the wallet, go to the send coins page and then change the coin to Testcoin. What happens is that the balance takes a long time to update (it is updated only in the next execution of `WalletService.loadBalances`), but when returning to Skycoin the balance is refreshed normally. The problem can be solved simply by changing `this.addresses.subscribe` to `this.addresses.first().subscribe` in `WalletService.loadBalances` and calling `WalletService.loadBalances` manually in `HeaderComponent` when the user changes the coin. Even without the problem this would be a good change, because over time the complexity of the subscriptions is becoming greater and there are several that are leaking, because they are not cleaned when appropriate.

The other part is that changing the coin should not only update the balance, but should check the status of the node first, such as when the wallet has just been opened. This is because each coin has a different node and one o those nodes could be out of service or out of sync.

AndreiIarovoi:
Fixes #351

Changes:
- problem with balance loading was fixed
- check the block chain after changing the current coin
- unit test fixed

### DIFF
--- a/src/app/components/layout/header/header.component.ts
+++ b/src/app/components/layout/header/header.component.ts
@@ -29,7 +29,7 @@ export class HeaderComponent implements OnInit, OnDestroy {
   highest: number;
   currentCoin: BaseCoin;
 
-  private isBlockchainLoading = false;
+  private isBlockchainLoading = true;
   private isBalanceLoaded = false;
   private price: number;
   private priceSubscription: Subscription;
@@ -51,8 +51,8 @@ export class HeaderComponent implements OnInit, OnDestroy {
   ngOnInit() {
     this.coinSubscription = this.coinService.currentCoin
       .subscribe((coin: BaseCoin) => {
+        this.reloadBlockchain();
         this.reloadBalance();
-        this.blockchainService.loadBlockchainBlocks();
         this.currentCoin = coin;
       });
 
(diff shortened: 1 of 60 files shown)


src/app/components/layout/header/header.component.ts
```diff
@@ -91,11 +92,13 @@ export class HeaderComponent implements OnInit, OnDestroy {
     this.coinSubscription.unsubscribe();
   }
 
-  private resetBalance() {
+  private reloadBalance() {
     this.coins = 0;
     this.price = null;
     this.balance = null;
     this.isBalanceLoaded = false;
+
```
Senyoret1: Please, add this to avoid some errors in the UI:

```
this.isBlockchainLoading = true;
this.percentage = null;
this.current = null;
this.highest = null;
```

Senyoret1:
Travis refuses to approve this. It may be a good idea to increase the time limit for obtaining the price, in the e2e test.

AndreiIarovoi:
We are trying to fix it now. Please, wait with the merge.

alexandermushinskiy:
During fixing the test `should display price information` we have found some issues with displaying the balance when there are no active connections. After solution has been found there are no ability to check the price information due to the backend for e2e tests is isolated and as a result we don't have any active peers. In this case the `Loading ...` message is shown instead of the balance and there is no way to check the price.
We have several solutions:
- the simplest one is to remove that test (`should display price information`);
- display the balance in case of there are no any active connections (there is a possibility to show unreliable balance).

@Senyoret1, which of these solutions is more preferable?

Senyoret1:
For now, just delete it. I have to check if the error messages are adequate for the new multicoin functionalities, I will write down the error to see if any solution appears during those tests. Additionally, please solve the conflict.

alexandermushinskiy:
Sure, the conflicts will be resolved. Thanks for the quick response.

PR #372, opened 2018-06-26
#366: Small improvements for the code

### LINKED ISSUE #366: Small improvements for the code
- `WalletService.isWalletsExist` is not a very descriptive name. It should be changed to `haveWallets`.

- The `defaultCoin` variable should be removed from `OnboardingCreateWalletComponent`. After removing the variable, make `OnboardingCreateWalletComponent.initForm` obtain the data like in `CreateWalletComponent.initForm`.

AndreiIarovoi:
Fixes #366 

Changes:
- Small improvements for the code have been done


Senyoret1:
This PR has several conflicts and problems due to the PRs that were previously merged:

- The renaming of `WalletService.isWalletsExist` will have problems with the `test-mocks.ts` added in https://github.com/skycoin/skycoin-web/pull/373. Now the name must be replaced in `test-mocks.ts` too.

- `defaultCoin` was replaced with `currentCoin` in https://github.com/skycoin/skycoin-web/pull/370, so there is no need to remove that variable in `OnboardingCreateWalletComponent`. In fact, now the problem is that it is not necessary to send `currentCoin` to `initForm()` in `CreateWalletComponent`.

If it is more comfortable, the modifications could be made in another PR.

PR #380, opened 2018-06-27
#367: Save the selected coin in every tab

### LINKED ISSUE #367: Save the selected coin in every tab
If a user opens 2 tabs, changes the coin in tab B and then refreshes the page in tab A, then tab A will show the same coin as tab B. This should not happen, a user should be able to open a tab for each coin and there should be no problems when refreshing the page.

To solve this, `CoinService` should save the current selection not only in `localStorage`, but also in `sessionStorage`. Additionally, when loading the current selection in `CoinService.loadCurrentCoin`, the value stored in `sessionStorage` should have priority, if it exists.

AndreiIarovoi:
Fixes #367 

Changes:
- currently selected coin is being stored in the sessionStorage as well in order to be able to work with different coins in the separate tabs

### DIFF
--- a/src/app/services/coin.service.ts
+++ b/src/app/services/coin.service.ts
@@ -15,6 +15,7 @@ export class CoinService {
   constructor() {
     this.loadAvailableCoins();
     this.loadCurrentCoin();
+    this.saveCoin(this.currentCoin.getValue().id);
   }
 
   changeCoin(coin: BaseCoin) {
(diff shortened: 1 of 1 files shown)


src/app/services/coin.service.ts
```diff
@@ -15,6 +15,7 @@ export class CoinService {
   constructor() {
     this.loadAvailableCoins();
     this.loadCurrentCoin();
+    this.saveCoin(this.currentCoin.getValue().id);
```
Senyoret1: Please use `sessionStorage.setItem(this.storageKey, this.currentCoin.getValue().id.toString());` instead of calling `saveCoin`, because it changes the selected coin in `localStorage` every time a tab is refreshed (assuming that the user has several tabs with different coins).

PR #383, opened 2018-06-28
#376: Refresh faster if the node is almost synchronized

### LINKED ISSUE #376: Refresh faster if the node is almost synchronized
Currently, when starting the wallet the code verifies if the node is synchronized and, if it is not, the verification is done again after 90 seconds. However, it is not too uncommon to see that the node needs only one block to be completely synchronized, case in which 90 seconds is too much. These cases usually happen when a transaction has recently been made and is still not fully replicated, so the node is not re-downloading the blockchain due to a problem.

If the node only needs 5 blocks or less, the verification should be done again after 5 seconds, instead of 90.

alexandermushinskiy:
Fixes #376

Changes:
- change interval period if the node is almost synchronized

### DIFF
--- a/src/app/services/blockchain.service.ts
+++ b/src/app/services/blockchain.service.ts
@@ -14,7 +14,9 @@ export class BlockchainService {
   private progressSubject: BehaviorSubject<any> = new BehaviorSubject<any>(null);
   private isLoaded = false;
   private intervalSubscription: Subscription;
-  private intervalPeriod = 90000;
+  private readonly defaultPeriod = 90000;
+  private readonly fastPeriod = 5000;
+  private intervalPeriod = this.defaultPeriod;
 
   get progress() {
     return this.progressSubject.asObservable();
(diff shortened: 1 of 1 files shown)


Senyoret1:
It is still updating after 90 seconds. The problem is that the interval does not change after modifying the value sent to `create()`. For testing, I changed the `onBlockchainProgress` code with this:

```
private temp = -1;
private onBlockchainProgress(response: any) {
  this.ngZone.run(() => {
    if (this.temp == -1) this.temp = 3;
    alert(this.temp);
    response.current -= this.temp;
    this.progressSubject.next(response);
    this.temp -= 1;

    this.calculateIntervalPeriod(response);

    if (response.current === response.highest) {
      this.completeLoading();
    }
  });
}
```

With that code the wallet should show how a block is added every 5 seconds, but it does not happen. In fact, if you modify `calculateIntervalPeriod` so that `intervalPeriod` is always set to 1000, the delay is still 90 seconds.

alexandermushinskiy:
We found solution, thanks for your example.

PR #399, opened 2018-07-02
#396: UI Fixes

### LINKED ISSUE #396: UI Fixes
- The header of the wallet list says “Encryption”, but it should say “locked”, because what the icon indicates, unlike in the desktop wallet, is if the wallet is locked and the seed must be introduced before being able to perform some actions.

- In the history page the amounts must be aligned to the right and the area for the amounts must be larger, as was done in https://github.com/skycoin/skycoin/pull/1683

- In the blockchain page, “Time since last block” should be replaced by “Timestamp of last block:”, since it is not showing the amount of time that has passed since the creation of the last block, but the date of the last block.

mahansky:
Fixes #396

Changes:
- "Encryption" renamed to "Locked"
- "Time since last block" replaced by "Timestamp of last block:"
- transaction history amount aligned to right

### DIFF
--- a/src/app/components/pages/history/history.component.scss
+++ b/src/app/components/pages/history/history.component.scss
@@ -91,7 +91,8 @@
       color: rgba(30, 34, 39, 0.5);
       display: inline-block;
       font-size: 13px;
-      margin: 4px 10px 0;
+      line-height: 15px;
+      margin: 0 4px;
     }
   }
 
--- a/src/app/components/pages/history/history.component.scss
+++ b/src/app/components/pages/history/history.component.scss
@@ -91,13 +91,13 @@
       color: rgba(30, 34, 39, 0.5);
       display: inline-block;
       font-size: 13px;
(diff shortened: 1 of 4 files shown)


src/app/components/pages/history/history.component.scss
```diff
@@ -91,13 +91,13 @@
       color: rgba(30, 34, 39, 0.5);
       display: inline-block;
       font-size: 13px;
-      line-height: 15px;
-      margin: 0 4px;
+      margin: 4px 10px 0;
```
Senyoret1: What is this change for? With this the address and the QR code button are not aligned:
al
Maybe it's better to modify `p` and `img` so that they are like in the desktop wallet.

mahansky:
I rolled-back the change for now, added a note to #400
It's doing oposite thing in my case

Original
screen shot 2018-07-02 at 16 56 12
With changes
screen shot 2018-07-02 at 16 55 42

PR #403, opened 2018-07-03
Add Spanish translation

Senyoret1:
Changes:
- `es.json` was added, with the texts in Spanish, and the language was added to `config.languages`.
- The `top-bar.*` text are now inside `header`, in the languaje files.
- A parameter was added to some text variables, in order to get the translated text with a single call to the `translate` pipe, instead of having to build the text by concatenating several strings.
- The "loading" texts now have 3 points at the end, instead of 2.

### DIFF
--- a/src/app/components/layout/header/top-bar/top-bar.component.html
+++ b/src/app/components/layout/header/top-bar/top-bar.component.html
@@ -1,6 +1,6 @@
 <div class="buttons-left">
   <span class="last-updated-time" *ngIf="isBalanceObtained">
-    {{ 'header.updated1' | translate }} {{ timeSinceLastUpdateBalances === 0 ? ('top-bar.less-than' | translate ) : timeSinceLastUpdateBalances }} {{ 'header.updated2' | translate }}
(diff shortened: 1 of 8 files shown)


src/app/components/layout/header/top-bar/top-bar.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="buttons-left">
   <span class="last-updated-time" *ngIf="isBalanceObtained">
-    {{ 'header.updated1' | translate }} {{ timeSinceLastUpdateBalances === 0 ? ('top-bar.less-than' | translate ) : timeSinceLastUpdateBalances }} {{ 'header.updated2' | translate }}
+    {{ 'header.top-bar.updated1' | translate }} {{ timeSinceLastUpdateBalances === 0 ? ('header.top-bar.less-than' | translate ) : timeSinceLastUpdateBalances }} {{ 'header.top-bar.updated2' | translate }}
     <i *ngIf="!this.isBalanceUpdated"
        class="material-icons blinking-warning-icon"
        [matTooltip]="'top-bar.connection-error-tooltip' | translate">
```
mahansky: `top-bar.connection-error-tooltip` should be `header.top-bar.connection-error-tooltip`

PR #405, opened 2018-07-05
Improve header images

Senyoret1:
Changes:
- Now the header images are in the `/coins` folder, inside `/assets/img`. Since each coin could have a different header image, this will help to maintain order after adding many more coins.

- The header images are now optimized 1280x720 .jpg files, instead of .png files. With this change, the Skycoin image went from 898KB to 89.3KB, and the Testcoin image from 1.79MB a 228KB. This should significantly improve the load times, especially on slow internet connections.

- The gradient shown in front of the header background is now not a CSS style, but an image. The name of the gradient image can be defined in the file of each coin (files inside `/src/app/coins`), so that now each coin can have a personalized header image and an adecuate gradient in front of it (previously the blue gradient of Skycoin was always shown).

- The text with the amount of time that has passed since the balance was updated is now slightly transparent, instead of slightly blue, so now it have fewer problems when changing the background image.


mahansky:
Is there any particular reason to have the gradient as a separate image? The effect could be added on top of the existing header image while saving a few extra kilobytes.

Senyoret1:
It is so that the design is more flexible and easier to personalize. With a bitmap you can put more complex effects easily, like a blur:
blur
That could be useful in designs like the one used by MDL:
mdl
Also, the process of configuring each coin with css styles is more complex, especially if the style has to be assembled by a designer without knowledge of css. With bitmaps he/she simply have to do tests in some image editor.

Senyoret1:
But, thinking about it, we could add the two options.

mahansky:
I mean you would have just one image with both the base texture and
gradient/blur stuff on top of it, all within same bitmap.

Senyoret1:
That was not possible before (because there was a fixed blue gradient in a css style), but with the changes in this PR you can simply keep `gradientName` undefined and then the wallet will only show the background image. After completing the multi-coin compatibility it will be necessary to create a file explaining how to add coins, including details like that.

However, the gradient was not originally added directly in the background bitmap because the visible area of the image varies depending on the width of the browser window. If the styles are modified so that the wallet looks good on smaller screens, then the visible area will vary a lot and it will be difficult to add directly in the background bitmap a gradient that looks good on all screens, but placing the gradient on top guarantee that at least the gradient will always be completely visible. Possibly some designs will need that and others will not.

Senyoret1:
The conflicts were fixed.

PR #406, opened 2018-07-05
Code cleaning - ApiService

Senyoret1:
Changes in the first commit:
- `Http` has been replaced with `HttpClient`, as `Http` has ben deprecated. The change was also made in `PriceService` to completely delete `Http`.
- The querystrings are created using `HttpParams`, instead of manually adding the text to the URL. 
- The type was added to various parameters and functions that did not have it.
- The function for processing the errors now returns the `error.error` text by default, because before it returned the complete error and the text of `error.message` was being used by default by the components, which was not adequate. For example, this error was being displayed:
lg
with the change now this error is shown:
imagen
- Small modifications were made to the e2e tests because they were failing locally.

In general, this pr was made while cleaning the `ApiService` code. There are still some methods that should not be inside `ApiService`, but they will be moved when cleaning the code of the services in which those methods should be.

### DIFF
--- a/src/app/services/api.service.ts
+++ b/src/app/services/api.service.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { Headers, Http, RequestOptions } from '@angular/http';
+import { HttpClient, HttpHeaders, HttpParams } from '@angular/common/http';
 import { Observable } from 'rxjs/Observable';
 import 'rxjs/add/observable/of';
 import 'rxjs/add/observable/throw';
@@ -16,7 +16,7 @@ export class ApiService {
 
   private url: string;
 
-  constructor(private http: Http,
+  constructor(private http: HttpClient,
               private translate: TranslateService,
               private coinService: CoinService) {
     this.coinService.currentCoin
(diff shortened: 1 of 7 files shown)


Senyoret1:
Changes in the second commit: 
- Now `ApiService.getErrorMessage` only returns text strings. The texts are filtered with `parseResponseMessage` directly in `ApiService.getErrorMessage`, so it is not longer needed to do it in many parts of the code before displaying the messages.

src/app/services/api.service.ts
```diff
@@ -41,60 +42,63 @@ export class ApiService {
...
-      : this.translate.get('service.api.server-error')
-          .flatMap(message => Observable.throw(new Error(message)));
+  private getErrorMessage(error: any): Observable<string> {
+    if (error.error) {
+      return Observable.throw(new Error(parseResponseMessage(error.error.trim())));
+    } if (error.message) {
+      return Observable.throw(new Error(parseResponseMessage(error.message.trim())));
+    }
+
+    this.translate.get('service.api.server-error')
```
mahansky: Missing `return`

src/app/services/price.service.ts
```diff
@@ -36,7 +36,7 @@ export class PriceService {
     }
 
     this.http.get(`https://api.coinmarketcap.com/v2/ticker/${this.cmcTickerId}/`)
-      .map(response => response.json())
-      .subscribe(response => this.price.next(response.data.quotes.USD.price));
+      .map(response => response)
```
mahansky: This line is not required anymore

src/app/services/api.service.ts
```diff
@@ -41,60 +42,63 @@ export class ApiService {
   }
 
   get(url, params = null, options = {}): Observable<any> {
-    return this.http.get(this.getUrl(url, params), this.returnRequestOptions(options))
-      .map((res: any) => res.json())
+    return this.http.get(this.getUrl(url), this.getRequestOptions(options, params))
+      .map((res: any) => res)
```
mahansky: not required

src/app/services/api.service.ts
```diff
@@ -41,60 +42,63 @@ export class ApiService {
...
   }
 
-  post(url, body = {}, options: any = {}) {
+  post(url, body = {}, options: any = {}): Observable<any> {
     return this.getCsrf().first().flatMap(csrf => {
       options.csrf = csrf;
-      return this.http.post(this.getUrl(url), body, this.returnRequestOptions(options))
-        .map((res: any) => res.json())
+      return this.http.post(this.getUrl(url), body, this.getRequestOptions(options))
+        .map((res: any) => res)
```
mahansky: not required

PR #411, opened 2018-07-12
Add the coin selection screen

### LINKED ISSUE #369: Design for choosing coins
Note: this design should NOT be implemented immediately, I put it here to get opinions.

Currently the wallet allows to select different coins, but the design is very simple. To change the coin, simply use the `<select>` in the upper right corner:
ac1
I did some tests (in an image editor, there is still no written code) and I think the appearance would improve using this style (expand the image to see the details):
ds1
It is a relatively simple change, which does not break the current graphic style. Previously I was testing another design, I do not like it that much (because it is less consistent with the current design and would be more difficult to integrate it with the desktop wallet later, if required), but I put it here just in case:
ds2
Additionally, I think it would be good to show the coin list in a modal window, like this one:
ds3
Any opinion is welcome.

Ping: @gz-c

Senyoret1:
Fixes #369 

Changes:
- A new button for changing the current coin was added to the header.
- The coin selection screen was created.
- The options for changing the coin that were present previously have not been eliminated yet.


Senyoret1:
:astonished: I changed it, thanks.

PR #414, opened 2018-07-13
 Language selection options 

Senyoret1:
Changes:
- The modal window for selecting the language was added:
modal

- The window is displayed when opening the wizard, before the the disclaimer, if the user does not have wallets.

- It is possible to change the language in one of the header menus. Probably it would be good to eliminate one of the buttons in another PR and keep only one menu, because the current distribution is a bit confusing: 
menu

- The black bar with options at the top of the header was deleted, because it is no longer necessary.

- The Test language was removed.

- The necessary additional e2e tests were created in the onboarding screen, so that the rest of the tests could work (due to the changes in the flow of the UI).


e2e/onboarding-create.e2e-spec.ts
```diff
@@ -20,6 +20,22 @@ describe('Onboarding Create', () => {
...
+  it('should not close the language selection modal by pressing outside the window', () => {
+    expect<any>(page.closeOutsideSelectLanguageModal()).toEqual(true);
+  });
+
+  it('should not close the language selection modal by pressing the X button', () => {
+    expect<any>(page.closeSelectLanguageModal()).toEqual(true);
+  });
+
+  it('should close the language selection modal by selecting a language', () => {
+    expect<any>(page.selectLanguaje()).toEqual(false);
```
mahansky: typo in method name

mahansky:
Here is an example of how it would look on Chrome

screen shot 2018-07-16 at 15 46 29

Senyoret1:
That sounds good. Do you have any particular font or library in mind?

mahansky:
You don't need a lib/font, it's just unicode characters. You can see them here with their codes http://www.unicode.org/emoji/charts/full-emoji-list.html#country-flag

For example code like this:
```
U+1F1E6 U+1F1EC
```
Can be converted to HTML by replacing `U+` with `&#x` and appending with `;` like this:
```
&#x1F1E7;&#x1F1EC;
```
Which will give you 🇧🇬

Senyoret1:
I did some tests and it probably would not be practical to use Unicode values due to compatibility issues. In a pc with Windows 10 the flags were not shown (although other emojis had no problems) in any of the browsers (latest versions of Chrome, Firefox and Edge). However, I was able to see the flags using Firefox on a PC with Windows 7, and on that same PC I could not see neither the flags nor the emojis using the lastest version of Chrome.

Maybe using bitmaps is the most consistent solution, plus it is easier to avoid copyright problems.

PR #417, opened 2018-07-14
Fix send tx -> change coin UX

Senyoret1:
Changes:
- If the user changes the coin while viewing the send coins form, the form is restarted, even if the system is in the process of creating the transaction (just before showing the preview).

- If the user changes the coin while viewing the preview of a transaction, the preview is closed and the send coins form is displayed again (without data).

- If the user tries to change the coin while a transaction is being sent, an error message is displayed:
msg


mahansky:
What is the `restarting` variable for?
https://github.com/skycoin/skycoin-web/pull/417/files#diff-5351343b12a6947e88a5d998ea493d85R43
I tried to remove it and I haven't noticed any difference in behaviour of the form.

Senyoret1:
It is used so that the form disappears for 1 frame. The goal is to avoid 2 problems:

- If you fill out the form and then change the coin, removing the form for a frame causes the fields to be reset.

- If you press the preview button and change the coin before the preview is shown, the form is reset correctly, but when the transaction is created the event is dispatched and the preview of the transaction created with the previous coin is displayed. Removing the form for a frame prevents that.

PR #434, opened 2018-07-18
Change tx icons

### LINKED ISSUE #422: Change the "sent" and "received" icons
The icons are confusing on the history page, because it is not immediately obvious which direction indicates if the transaction is incoming or outgoing.

Senyoret1:
Fixes #422 

Changes:
- The history page now shows a green icon for incoming transactions and a red one for outgoing transactions. Additionally, the icon is the same one used in the header, to maintain consistency. The change does not have esthetic ends, the objetive is to help to be able to quickly navigate through the history of transactions, because the icons that have been used until now are difficult to differentiate in a quick glance:
old
new

- `TransactionInfoComponent` uses the new colored icons, but only if opened in the history page. When sending a transaction the icon is blue, to avoid generating a negative feeling with the red color. However, the blue icon was replaced by the one in the header, to maintain consistency.


mahansky (approved):
Good idea, I'm not a huge fan of the colors though. I would probably try grey (sent) and blue (received) and leave TransactionInfoComponent without change. It's up to you..

Senyoret1:
I tried some other designs and I liked almost nothing, but maybe this could work:
icons
The vertical line allows to know if the arrow is entering or leaving, and the two designs seem to be different enough to differentiate them in a quick view of the list (mainly due to the overall shape). it's a very simple change, but at least it's consistent with the rest of the design. What do you think, @mahansky?

mahansky:
I don't know, I'm not too happy with that either. I mean, the logic behind it is good; Maybe a dot instead of a line with a bit of space? `. <` and `. >`.

Senyoret1:
I changed the lines with dots, as you suggested. Viewing the result I think it is not so easy to differentiate the two icons, as with the lines, which is a problem. Here is an example:
dot

I also modified the line icons, using smaller arrows. I think it looks a little better than with the big arrows:
small

Actually I did some tests with more elaborate designs, but the design of the wallet has very simple elements. It is difficult to use an icon with a complex design and make it look consistent, so I think the options are few. Here is an example of a complex design I tried:
shadow

mahansky:
The last one is pretty sweet but as you said, it doesn't really fit into the current design.
What if you combined the lines and that last design in such a way that the line would be on the very left of the transaction row (similar to border but on the inside and not full height) and the arrow the way it was or maybe smaller, bit closer to the left?
Kind of hard to describe it but I dont have any img editor on this computer.

gz-c:
@Senyoret1 this is very old, are we continuing with it?

Senyoret1:
@gz-c I stoppend working on this long ago, so I will close it. If needed, I will check it again later

PR #442, opened 2018-07-19
Use only one menu

### LINKED ISSUE #438: Improve the header menus
Currently it does not seem necessary to have more than one menu in the upper right corner. Additionally, the icons do not seem to indicate clearly what each menu is for.

Senyoret1:
Fixes #438 

Changes:
- The two menus in the upper right corner were merged into one.

- The menu button is now a little more to the right (an amount of space equal to the left margin of the button for changing the coin), so that it is possible to see the right edge of the menu.
result


mahansky:
When I removed the width of the button (50 to auto/24) and added a right margin (10) instead, it created the desired effect while keeping the margins same on the both sides of the header.
screen shot 2018-07-23 at 14 13 10

Senyoret1:
I tried what you said, but I could not make the button highlighting work well, so I used a dummy object.

PR #446, opened 2018-07-21
Improvements around WalletDetailComponent

Senyoret1:
Changes:
- When the user has already added 5 addresses or more and tries to add a new one, a message is displayed telling that the wallet will take longer to be unlocked.
- Now the addresses in `WalletDetailComponent` are copied using `ClipboardDirective`, as it is done in the QR code modal window.
- Now the texts for the confirmation dialogs are obtained in a simpler way, in `WalletDetailComponent`.
- `ConfirmationData` was optimized, because not all data is always necessary.
- Other improvements for the code.


src/assets/i18n/en.json
```diff
@@ -131,6 +131,7 @@
     "delete-confirmation1": "WARNING: the wallet",
     "delete-confirmation2": "will be deleted. If you forget your seed, you will not be able to recover this wallet! Do you want to continue?",
     "delete-confirmation-check": "Yeah, I want to delete the wallet.",
+    "add-confirmation": "If you add more addresses, the process of unlocking the wallet to perform certain operations will take more time. do you want to continue?",
```
mahansky: Capital letter: **Do** you want to continue?

PR #447, opened 2018-07-24
Improved code for the wallet creation forms

Senyoret1:
Changes:
- Now the seeds are checked before creating the wallets. The objective is to alert the user if the seed does not correspond to the characteristics of the seeds created automatically by the system, to avoid simple mistakes that could prevent recovering the wallet in the future. If the validation fails, this message is displayed and the user must confirm if really wants to use the seed:
confirmation

- Now the form used in the modal window and in the wizard is in a reusable component, to avoid repeating code.

- Improvements to make the wizard look better on small screens.

- Some improvements to the code.


Senyoret1:
The problems with Travis were solved.

src/app/components/pages/onboarding/onboarding-create-wallet/onboarding-create-wallet.component.ts
```diff
@@ -13,60 +11,41 @@ import { CoinService } from '../../../../services/coin.service';
...
 export class OnboardingCreateWalletComponent implements OnInit {
+  @ViewChild('formControl') formControl: CreateWalletFormComponent;
   @ViewChild('create') createButton;
 
   showNewForm = true;
-  form: FormGroup;
   doubleButtonActive = DoubleButtonActive.LeftButton;
-  haveWallets = false;
-  isWalletCreating = false;
+  userHaveWallets = false;
```
mahansky: user has wallets

PR #448, opened 2018-07-26
Improvements for the multi-coin options

### LINKED ISSUE #444: Improvements for BaseCoin and other related classes
- `nodeVersion` should not be required as part of `BaseCoin`. The code should get that info automatically when the initial check is made before checking if the node is synchronized.

- `cmcTickerId` should be optional. Changes will have to be made in other parts of the code, to avoid problems after doing that.

- `coinExplorer` should be optional.

- `imageName` should be optional. It will be necessary to create a generic image as fallback.

- `CoinService` should detect if there is more than one coin with the same id.

Senyoret1:
Fixes #444 

Changes:
- The version of the node is obtained automatically by the code.

- `cmcTickerId`, `coinExplorer` and `imageName` are now optional.

- `CoinService` detects if there is more than one coin with the same id.

- The documentation was updated.


src/app/services/blockchain.service.ts
```diff
@@ -109,10 +111,12 @@ export class BlockchainService {
       .map((status: any) => {
         if (!status.connections || status.connections.length === 0) {
           this.onLoadBlockchainError(ConnectionError.NO_ACTIVE_CONNECTIONS);
-          return null;
+          throw { reported: true };
         }
 
         return status;
```
mahansky: this return is no longer necessary

src/app/services/blockchain.service.ts
```diff
@@ -109,10 +111,12 @@ export class BlockchainService {
...
-      });
+      })
+      .flatMap(() => this.apiService.get('version'))
```
mahansky: this wont happen if the exception above is thrown, but I guess it's ok since the nodeVersion is not used anywhere yet

src/app/components/layout/header/header.component.html
```diff
@@ -1,16 +1,18 @@
 <div class="-container">
   <app-disclaimer-warning></app-disclaimer-warning>
 
-  <div class="large-header" [ngStyle]="{'background-image': 'url(../../../../assets/img/coins/' + currentCoin.imageName + ')'}">
-    <div class="gradient" [style.background-image]="currentCoin.gradientName ? 'url(../../../../assets/img/coins/' + currentCoin.gradientName + ')' : 'none'">
+  <div class="large-header" [ngStyle]="{'background-image': 'url(assets/img/coins/' + currentCoin.imageName + ')'}">
```
mahansky: this should probably check if the imageName is set

src/app/services/blockchain.service.ts
```diff
@@ -66,7 +68,7 @@ export class BlockchainService {
       .filter(status => !!status)
```
mahansky: this filter should probably be removed as the status is no longer related to connections but version

src/app/services/blockchain.service.ts
Senyoret1: At this time, when the node does not have connections to other nodes, an error message is displayed and the entire wallet is stopped, so `nodeVersion` would not be needed (even after starting to use it). The wallet needs something to allow checking the node status again, but in any case the error should be very rare.

PR #459, opened 2018-08-09
Start integrating with Electron

### LINKED ISSUE #457: Start integrating with Electron

Senyoret1:
Fixes #457 

Changes:
- First steps to pack the wallet with Electron. The integration code is based on the one used for the desktop wallet, but it has some differences (mainly because the contents are not loaded from a server and that there is no need to run a node).

It is necessary to make deeper revisions. A quick error to detect is that the wallet does not show the white background of `SelectCoinOverlayComponent`. It is also not possible to send coins.

To be able to try the wallet it is necessary to put `nodeUrl = environment.production ? 'http://127.0.0.1:6420/api/v1/' : '/api/v1/';` in https://github.com/skycoin/skycoin-web/blob/f4e57e705520c5853ad80d98245f78564e3774d4/src/app/coins/skycoin.coin.ts#L7


gz-c:
Is the node url configurable by the user?

Senyoret1:
No, the URLs of the nodes are assigned in the code files that contains the configuration of each coin:
https://github.com/skycoin/skycoin-web/tree/develop/src/app/coins

mahansky:
On top of those two errors you mentioned:
- there is the same problem with (some) https requests, electron cancelling them with "INSECURE_RESPONSE"
- refreshing page breaks the app because of the URL change (.../dist/wallets for example which doesnt exists)

gz-c:
We should have some kind of config panel to allow the user to set the backend url

Senyoret1:
Additional changes:
- A simple server was created to serve the wallet files (`electron/server.go`). This was made because the wallet was created using a server, so using local files cause problems. To compile the server you simply need to run `build-server.sh`.

- The problem that made imposible to refresh the page with f5 was solved.

- The hexadecimal colors used on the css files were replaced with `rgba()`, to avoid problems with some browsers and with Electron. This fixes the problem with the `SelectCoinOverlayComponent` background.

- Now the wallet can be connected to `node.skycoin.net`, so production builds work correctly.

- It is posible to send coins.

- Now the code does not request and send csrf tokens, because the public nodes must have them disabled. Requesting a csrf tokens from a node started with `-disable-csrf` returns an error.

- Added new commands to `Makefile`.

- Updated parts of the Readme.

I could not find "INSECURE_RESPONSE" errors, I do not know if they were solved by any of the changes made.

There are still problems and the screen for manually entering the node URL has not been added. However, the changes are not too complex. I am going to open issues with what is still pending.

Senyoret1:
I added a Readme, based in the one on the desktop wallet repo (https://github.com/skycoin/skycoin/blob/develop/electron/README.md)

Senyoret1:
Additional changes:
- A Readme file was included in the Electron folder
- A page that allows the user to change the node URL of each coin was added (it works both in the browser and in Electron)

@mahansky I think that with these changes the PR is ready for review.

mahansky:
The electron app won't start because it can't find server' executable on MacOS.
It looks for it here:
`electron/release/mac/Skycoin.app/Contents/MacOS/server`
but it's located in 
`electron/release/mac/Skycoin.app/Contents/server`
It's probably different on every platform, skycoin wallet deals with something similar this way https://github.com/skycoin/skycoin/blob/develop/electron/src/electron-main.js#L53

I also noticed that the `build-server.sh` script doesn't have exec permissions

Senyoret1:
I fixed the problem with the path of the server excecutable. I also fixed a problem on the Mac that made it impossible to reopen the app after closing it, unless the server was manually stopped. Additionally, I updated the Electron version and modified the permissions of `build-server.sh`.

In the tests I did the app started correctly on Mac and Linux. However, the development environment did not allow me to test if the server closes properly in Linux when the app is closed, but the code that makes it work on Mac should make it work correctly in Linux (I'll keep trying to test that).

electron/src/electron-main.js
```diff
@@ -0,0 +1,292 @@
...
+const serverPort= 8412;
+
+// It is only possible to make connections to hosts that are in this list.
+var allowedHosts = new Map();
+allowedHosts.set('127.0.0.1:' + serverPort, true);
+allowedHosts.set('api.coinmarketcap.com', true);
+allowedHosts.set('api.github.com', true);
+
+// Indicates if the URLs of the nodes were already added to allowedHosts.
+var nodesAllowed = false;
```
mahansky: this var is never `true`

electron/src/electron-main.js
```diff
@@ -0,0 +1,292 @@
...
+
+  if (server) {
+    console.log('Server already running');
+    return
+  }
+
+  // Resolve server binary location
+  var exePath = path.dirname(app.getPath('exe'));
+  var exe = (() => {
+    switch (process.platform) {
```
mahansky: you can merge one branch with default in both `switch`

electron/src/electron-main.js
```diff
@@ -0,0 +1,292 @@
...
+        return path.join(exePath, './dist/');
+    }
+  })()
+
+  // Start the server
+  server = childProcess.spawn(exe, ['-port=' + serverPort, '-path=' + contentsPath]);
+
+  createWindow();
+
+  server.on('error', (e) => {
```
mahansky: I always end up adding `e` into `console.log` when debugging, it would be nice to have it there, also with the exit codes..

electron/src/electron-main.js
```diff
@@ -0,0 +1,292 @@
...
+
+  session
+    .fromPartition('')
+    .setPermissionRequestHandler((webContents, permission, callback) => {
+      return callback(false);
+    });
+
+  // Blocks the connection if the URL is not in allowedHosts.
+  session.defaultSession.webRequest.onBeforeRequest(['*://*./*'], function(details, callback) {
+    if (details.url.startsWith('http')) {
```
mahansky: is this intentional to filter protocol? because otherwise it works for both http and https

electron/src/electron-main.js
```diff
@@ -0,0 +1,292 @@
...
+      if (!nodesAllowed) {
+        allowNodes();
```
mahansky: `allowNodes` does async stuff, it should return the promise so the `callback` can be called when it's done

src/app/components/pages/settings/nodes/change-url/change-node-url.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  ngOnInit() {
+    this.initForm();
+  }
+
+  closePopup() {
+    this.dialogRef.close();
+  }
+
+  change() {
```
mahansky: this function (or coinservice) should probably update `nodeURLs` global var to allow requests to new url (I tried to update it to local node but it didn't work, also when changing nodeURLs manually, not sure what exactly was wrong)

mahansky:
Btw the script still doesn't have exec permissions. Try this command: `git update-index --chmod=+x build-server.sh`

Senyoret1:
I changed the way of passing data between the main process and the web app, so that it is now more flexible and avoids the problem of the asynchronic call. Also, `build-server.sh` should now work well.

electron/src/electron-main.js
Senyoret1: The variable was deleted because now the procedure is different

electron/src/electron-main.js
Senyoret1: I added `e ` to `console.log`, but I'm not sure if that's exactly what you had in mind

electron/src/electron-main.js
Senyoret1: Just old code that remained after several changes

electron/src/electron-main.js
Senyoret1: I removed the asynchronous procedure. Now the data is injected synchronously with the Electron ipc

src/app/components/pages/settings/nodes/change-url/change-node-url.component.ts
Senyoret1: Fixed. It probably won’t work when using a local node because the node rejects connections from different sources.

PR #461, opened 2018-08-14
General improvements

### LINKED ISSUE #460: Top right menu autofocus
The "Blockchain" item in the top right menu is always focused after opening it

Senyoret1:
Fixes #460 

Changes:
- The first item of the top right menu is no longer selected. The change also implies that the menu is again shown touching the right edge of the browser, since there is no way to set the position of the menu and the method that was previously used made the menu to always be displayed with the first item selected. The problem originated because the Material library always shows the first item selected if it does not detect real mouse interaction ( https://github.com/angular/material2/blob/master/src/lib/menu/menu-trigger.ts ), so the problem is too complex for the little importance of the matter.

- The amount of coins in the pending transactions page is now aligned to the right, to maintain consistency with the other pages.

- The text that indicates how much time has elapsed since the balance was updated has been changed, in order to make clearer what it refers to. Now, instead of "Updated:" it says "Balance updated:".

- An animation is now displayed while the balance is being updated. The animation is shown to the right of the text that indicates how much time has elapsed since the balance was updated.

- Now the code cleans the subscription for unlocking the wallet, because it could cause some problems when using the navigation buttons of the browser.

- Some minor changes in the code.

### DIFF
--- a/src/app/components/layout/header/top-bar/top-bar.component.html
+++ b/src/app/components/layout/header/top-bar/top-bar.component.html
@@ -2,9 +2,10 @@
   <app-coin-button *ngIf="hasManyCoins" (click)="changeCoin()"></app-coin-button>
   <div class="last-updated-time" *ngIf="balanceObtained">
     <div class="label">{{ 'header.top-bar.updated1' | translate }}</div>
-    <div>
+    <div class="header-sel-theme">
       {{ timeSinceLastBalanceUpdate === 0 ? ('header.top-bar.less-than' | translate ) : timeSinceLastBalanceUpdate }} {{ 'header.top-bar.updated2' | translate }}
-      <i *ngIf="!this.problemUpdatingBalance"
+      <mat-spinner *ngIf="updatingBalance" color="primary" class="in-button very-small loading"></mat-spinner>
+      <i *ngIf="this.problemUpdatingBalance && !updatingBalance"
         class="material-icons blinking-warning-icon"
         [matTooltip]="'header.top-bar.connection-error-tooltip' | translate">
         warning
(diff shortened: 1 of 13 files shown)


src/app/components/layout/header/top-bar/top-bar.component.html
```diff
@@ -2,9 +2,10 @@
   <app-coin-button *ngIf="hasManyCoins" (click)="changeCoin()"></app-coin-button>
   <div class="last-updated-time" *ngIf="balanceObtained">
     <div class="label">{{ 'header.top-bar.updated1' | translate }}</div>
-    <div>
+    <div class="header-sel-theme">
       {{ timeSinceLastBalanceUpdate === 0 ? ('header.top-bar.less-than' | translate ) : timeSinceLastBalanceUpdate }} {{ 'header.top-bar.updated2' | translate }}
-      <i *ngIf="!this.problemUpdatingBalance"
+      <mat-spinner *ngIf="updatingBalance" color="primary" class="in-button very-small loading"></mat-spinner>
```
mahansky: Spinner never shows up because it is inside `*ngIf=balanceObtained`

src/app/components/layout/header/top-bar/top-bar.component.ts
```diff
@@ -54,13 +53,13 @@ export class TopBarComponent implements OnInit, OnDestroy {
     );
 
     this.subscription.add(
-      this.balanceService.totalBalance.subscribe((balance: TotalBalance) => {
-        if (balance && !this.balanceObtained) {
-          this.balanceObtained = true;
+      this.balanceService.totalBalance.subscribe(balance => {
+        if (balance) {
+          this.updateTimeSinceLastBalanceUpdate();
+          this.balanceObtained = balance.state === BalanceStates.Obtained;
```
mahansky: when the balance is being updated, obtained is false, which makes the update animation hidden

Senyoret1:
Yeah, you are right. Additionally, while making the change I saw that the header was taking longer than it should to be updated after launching events with `BalanceService.totalBalance`, due to the `_ngZone.runOutsideAngular` in `BalanceService.scheduleUpdate`, so I solved that too.

Senyoret1:
Oh, I did not know the PR was on the fork. I'm going to merge this PR as it is because the changes to the menu button cause it to lost part of its clicqueable area I think it is possible to avoid that by a mix of how the menu was positioned before and avoiding the checks that Angular Material makes to know if it was actually opened with a real mouse click (Angular Material actually only wants to know if an accessibility tool is being used). I will try it later, but I'll keep the changes you made as backup in case it does not work.

PR #462, opened 2018-08-15
Prepare for a test server

Senyoret1:
In preparation for uploading the wallet to a server for testing, this PR disables Testcoin. This means that the options for changing the coin will be hidden in the UI, since only one coin would be available (Skycoin). It is possible to reactivate Testcoin simply by uncommenting a line of code, in case it is necessary for testing.

Update 1: now Testcoin is only hidden in production builds

Update 2: the style of the URLs was changed to `HashLocationStrategy`, to avoid having to make additional configurations to the server.

### DIFF
--- a/e2e/onboarding-create.e2e-spec.ts
+++ b/e2e/onboarding-create.e2e-spec.ts
@@ -118,7 +118,7 @@ describe('Onboarding Create', () => {
   });
 
   it('should load wallet with correct address', done => {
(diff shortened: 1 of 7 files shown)


gz-c:
Can we use a feature flag?

Senyoret1:
Additional change, indicated as "Update 2" in the first comment.

PR #472, opened 2018-10-13
Add a test env for Electron

Senyoret1:
This pr is based on https://github.com/skycoin/skycoin/pull/1933

Changes:
- In order to make development with Electron easier, this pr adds the command `npm start` inside the `electron` folder, to open a development Electron window that instead of starting a dev server and then showing the contents of the `dist` folder, it simply shows the contents of http://localhost:4200. In this way it is possible to simply modify the UI as while developing for the web and see the changes directly in the Electron window, without needing to recompile.
- Most of the changes were limited to placing pre-existing code inside `if (!dev)`, but because of the tabulation changes Github shows changes in a large number of lines.


Senyoret1:
Travis should stop failing after https://github.com/skycoin/skycoin-web/pull/474

PR #478, opened 2018-10-23
Remove yarn from the readme

Senyoret1:
Changes:
- Remove the references to Yarn from the readme. Yarn stopped being used in https://github.com/skycoin/skycoin-web/commit/a0776d97924d1a0479e33fa1bdf9d731274851c8

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,7 @@ The Skycoin web interface requires Node 6.9.0 or higher, together with NPM 3 or 
 
 This project is generated using Angular CLI, therefore it is adviced to first run `npm install -g @angular/cli`.
 
-Dependencies are managed with Yarn, to install yarn run `npm install -g yarn`.
-
-To install all Angular, Angular CLI and all other libraries, you will then have to run `yarn`.
+To install all Angular, Angular CLI and all other libraries, you will have to run `npm install`.
 
 You will only have to run this again, if any dependencies have been changed in the `package.json` file.
 


mahansky:
I noticed Yarn is still used in the Dockerfile, not sure if you want to create another PR or change it here https://github.com/skycoin/skycoin-web/blob/011c01b166f52073b12bd9b91b8c536859fdc53f/docker/images/Dockerfile#L8

Senyoret1:
Perhaps Yarn was used in the Dockerfile just following the outdated info of the readme, I'm not sure. @olemis, could you take a look at this?

olemis:
@stdevRob could you pls take a look into this ?

stdevYuniers:
yarn has been installed on node:8-alpine, therefore is not necessary do `npm install -g yarn` on README file. @Senyoret1 have reason with remove it from README file.

I recommend use yarn above npm, yarn is a best way to manage our node packages.

# It would be a nice solution

## Installation

This project is generated using Angular CLI, therefore it is adviced to first run `yarn global add @angular/cli`.

To install all Angular, Angular CLI and all other libraries, you will then have to run `yarn`.

You will only have to run this again, if any dependencies have been changed in the `package.json` file.

## Compiling new target files

To compile new target files, you will have to run: `yarn run build`

# And our [Dockerfile](docker/images/Dockerfile) can be changed to

```
RUN yarn global add @angular/cli \
    && yarn \
    && yarn run build
```

Senyoret1:
In fact, I removed Yarn from the readme not because it was not necessary for the docker image, but because it is not used in the normal development process. The repository does not have a lock file for Yarn so using Yarn or NPM is probably similar, although the repository does have a `package-lock.json` file. The desktop wallet also uses NPM instead of Yarn and there even was a problem sometime ago and we changed from Yarn to NPM for Electron dependencies https://github.com/skycoin/skycoin/issues/1733 . The change in this pr is mainly to indicate the normal development process that has been used up to now

stdevYuniers:
Maybe @samuelvisscher can say more arguments of use npm instead of yarn.

I'm not have problems with change it on README file. Docker build instructions can  remain the same

olemis:
@stdevYuniers please remove all references to `yarn` in the `Dockerfile` . Thanks @Senyoret1 for the explanation .

Senyoret1:
Added in https://github.com/skycoin/skycoin-web/pull/484

PR #485, opened 2018-10-28
Use v0.25.0 api

### LINKED ISSUE #160: Replace /explorer/address with /transactions
The transaction history is obtained usign `WalletService.transactions()` and `WalletService.retrieveAddressTransactions()`:
https://github.com/skycoin/skycoin-web/blob/6eedb6576b5572e1aa0d2e6fc6af2270682d19cb/src/app/services/wallet.service.ts#L150-L195

The problem is that the procedure checks all the addresses one by one, making as many connections as the user has addresses. As connections are made to a remote server, this has a significant overheat. To improve this, the procedure for obtaining transactions should use [/transactions](https://github.com/skycoin/skycoin/blob/develop/src/api/README.md#get-transactions-that-are-addresses-related) instead of [/explorer/address](https://github.com/skycoin/skycoin/blob/develop/src/api/README.md#get-address-affected-transactions), since `/transactions` allows to request transactions from several addresses at the same time. All transactions could be obtained with a procedure similar to that of `WalletService.retrieveAddressesBalance()`, where the balance of all the addresses is obtained at the same time:
https://github.com/skycoin/skycoin-web/blob/6eedb6576b5572e1aa0d2e6fc6af2270682d19cb/src/app/services/wallet.service.ts#L290-L293

While using `/transactions`, please make sure to ask for all transactions, by no adding the `confirmed` args.

Senyoret1:
Fixes #160 

Changes:
- The connections to the node now use many of the improvements added in v0.25.0, Mainly the `verbose` argument and the ability to send long lists of addresses using POST.
- A way to determine the version of the node was created. This is currently used to maintain compatibility with the version 0.24.0 of the node and to use the new features only if the node is runnin v0.25.0 or higher. 
- The function responsible for sending POST requests was updated, to be able to send more than Json.
- The unit tests were updated to be compatible with the changes.

### DIFF
--- a/src/app/services/api.service.ts
+++ b/src/app/services/api.service.ts
@@ -30,23 +30,12 @@ export class ApiService {
       .catch((error: any) => this.getErrorMessage(error));
   }
 
-  post(url, body = {}, options: any = {}, requestCsrf: boolean = false): Observable<any> {
-    if (requestCsrf) {
-      return this.getCsrf().first().flatMap(csrf => {
-        options.csrf = csrf;
-        return this.http.post(
-          this.getUrl(url),
-          options.json ? JSON.stringify(body) : this.getQueryString(body),
-          this.getRequestOptions(options)
-        ).catch((error: any) => this.getErrorMessage(error));
-      });
-    } else {
-      return this.http.post(
-        this.getUrl(url),
-        options.json ? JSON.stringify(body) : this.getQueryString(body),
-        this.getRequestOptions(options)
-      ).catch((error: any) => this.getErrorMessage(error));
-    }
+  post(url, body = {}, options: any = {}): Observable<any> {
+    return this.http.post(
+      this.getUrl(url),
+      options.json ? JSON.stringify(body) : this.getQueryString(body),
+      this.getRequestOptions(options)
+    ).catch((error: any) => this.getErrorMessage(error));
   }
 
   private getQueryString(parameters = null) {
--- a/src/app/services/api.service.ts
+++ b/src/app/services/api.service.ts
@@ -34,13 +34,31 @@ export class ApiService {
     if (requestCsrf) {
       return this.getCsrf().first().flatMap(csrf => {
         options.csrf = csrf;
-        return this.http.post(this.getUrl(url), body, this.getRequestOptions(options))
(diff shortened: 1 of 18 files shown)


src/app/services/api.service.ts
```diff
@@ -34,21 +34,39 @@ export class ApiService {
     if (requestCsrf) {
```
mahansky: `requestCsrf` is never `true`, this could be removed

Senyoret1:
Compatibility with previous versions is maintained for several reasons. One of the main ones is to make it easily to update the wallet and the public node without downtime: if the wallet is compatible with both versions, it is possible to update the wallet before the public node and everything will work correctly both before and after the public node is updated. Another reason, even more important, is that the wallet is compatible with multiple coins and all coins could be running a different version of the node (in fact in the future some coins could be several version behind compared to Skycoin).

As you suggested, I simplified the version checking code a bit. However, it is not optimal to simply save the node version in a variable and consult it, because the version of the node is obtained asynchronously after changing the coin or manually changing the URL of the node. The main problem is that when changing the coin the code starts to perform operations that need to know the version of the node, but the variable still contains the version of the node of the previously selected coin. The approach used in this pr allows to ensure that the version of the node is readed only after having updated it, to avoid the previous problem. Using `BehaviorSubject` with an initial value of null instead of `ReplaySubject` is part of the approach, since null is used to indicate that the version of the node has not yet been updated (the value returns to be null when needed, like after changing the coin).

I created a new issue with the parts that should use BigNumber: https://github.com/skycoin/skycoin-web/issues/486

I also deleted the code related to `requestCsrf`, as you recommended.

gz-c:
For version comparison, use the X.YY in vX.YY.ZZ, ZZ is for bugfix releases and won't affect the API

Senyoret1:
The code that makes the comparison takes into account the Z value, just in case. However, the code is only checking if the version is 0.25.0 or higher, so in practice the Z value is not being taking into account. The "rc" part is ignored.

gz-c:
The Z part needs to be ignored in case we have to do a bugfix release, we won't want to interfere with the light client behavior

Senyoret1:
I modified the function and now the Z part is always ignored

PR #496, opened 2018-12-09
Validate the correct number of decimals

Senyoret1:
Changes:
-  The form for sending coins no longer checks if the number of coins has 6 or fewer decimal places, but validates if the number of decimal places is not greater than the amount currently activated in the node. The number of valid decimal places is obtained with a call to the `health` api endpoint

### DIFF
--- a/src/app/components/pages/send-skycoin/send-form/send-form.component.ts
+++ b/src/app/components/pages/send-skycoin/send-form/send-form.component.ts
@@ -52,8 +52,6 @@ export class SendFormComponent implements OnInit, OnDestroy {
     this.subscription.add(this.coinService.currentCoin
       .subscribe((coin: BaseCoin) => this.currentCoin = coin)
     );
-
-    window['blockchainService'] = this.blockchainService;
   }
 
   ngOnDestroy() {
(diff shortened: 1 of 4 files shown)


src/app/components/pages/send-skycoin/send-form/send-form.component.ts
```diff
@@ -147,7 +153,7 @@ export class SendFormComponent implements OnInit, OnDestroy {
 
     const parts = amountControl.value.toString().split('.');
 
-    if (parts.length === 2 && parts[1].length > 6) {
+    if (parts.length === 2 && parts[1].length > window['blockchainService'].currentMaxDecimals) {
```
mahansky: What's the reason to store blockchainService in a window object?

Senyoret1: When `validateAmount` is called, `this` is `undefined`, so checking `this.blockchainService.currentMaxDecimals` causes an error. Using `window` to temporarily store `blockchainService` is just a quick way to fix the problem.

mahansky: It's probably caused by missing `.bind(this)` here 
https://github.com/skycoin/skycoin-web/blob/7879a58ef06f1905829693a44453231f6aa8a8cd/src/app/components/pages/send-skycoin/send-form/send-form.component.ts#L136

Senyoret1: You are right. I changed it, thanks.

PR #498, opened 2018-12-11
Add a button to update the balance

Senyoret1:
update

Changes:
- A button was added to allow to update the balance manually.
- A help message was added to inform the user that it is possible to update the balance manually. The purpose of the message is not only to inform what the button does, but also to ensure that the user know that the balance may take some time to be updated.


mahansky:
I feel like the popup (help message) isn't really necessary as it's just a simple action where tooltip would be sufficient. 
This looks like there is a full "app tour" - after you close one, another will show up with description of other button etc..

Senyoret1:
Yes, maybe it was a bit excessive. I removed it.

PR #502, opened 2018-12-28
Make the wallet work from the local file system

Senyoret1:
Changes:
- With this PR it is now possible to compile the web wallet and then open it from the local file system. To do so just run `make build-for-local-fs`.


Senyoret1:
This PR removes `npm build-for-electron` because it was not actually being used anywhere

PR #506, opened 2019-01-15
Responsive

### LINKED ISSUE #493: Make the design totally responsive
At this time the wallet requires the window to be at least 768 pixels wide. It is necessary to make the wallet work correctly on smaller screens, so that it works on mobile devices.

Senyoret1:
Fixes #493 

Changes:
- Now the design works well on small screens (tablets and cell phones).
- Now, when leaving the page, the user is asked for confirmation. This is especially useful to avoid problems with the back button of mobile devices.

I will test the change a little more, but there should not be many problems left.

There are some improvements that can be made, but I did not put them in this PR to not extend the development too much. The main ones are:

- It should be possible to close the modal windows with the return button of the telephone
- The performance of cryptographic functions seems to be very bad in low-end phones.
- Due to slow performance in mobile devices, there should be an animation when adding a new address to the wallet.


Senyoret1:
Another change is that it is now possible to access the test server from another device connected to the local network, simply by writing the ip address and the port (x.x.x.x:4200). However, it may be necessary to add rules to the firewall to make it work.

Senyoret1:
The last commit includes the following improvements:

- Fixed a problem added in the PR that stoped saving the data in the production builds. The problem was that after adding the first wallet on the wizard, any newly added wallet did not appear in the wallet list.

- A letter S that was appearing in the lower part of the system was removed (removed from `index.html`).

- When a cryptographic operation takes a long time to complete, a message is displayed indicating that the operation may take more than a minute on certain mobile devices and slow PCs. However, the message is only visible when the design for mobiles and tablets is shown, which are the devices on which the operation is most likely to be very slow. The message is displayed only if the system takes more than 7 seconds to complete a cryptographic operation.

- Small code improvements.

Senyoret1:
With the last commit a load animation is also displayed when a new address is being added and the device has a small screen.

Senyoret1:
The last commit contains the following changes:

- As it worked previously, the coin selection button is not displayed when there is only one coin.
- The zoom was blocked on mobile devices, to avoid problems with Safari.
- Now the wallet checks if the browser allows to access the `crypto` library inside a worker. If the check fails, an error message is displayed to the user stating that the wallet is not compatible with the browser.

With these changes I think I won't make additional modifications until the PR is reviewed. Additionally, it would be good if the PR could be check in modern versions of Safari, both in Mac OS and iOS, since I only have old versions that are not compatible with the `crypto` library.

Senyoret1:
Thanks for the review.

I changed the background of the modal windows to the same color used in the desktop wallet. I also fixed the problem with the global error messages and removed the autofocus on the modal windows that open the keyboard. I did not make changes about the last point because I don’t really know what would be the best thing to do, maybe it would be good to think a little more about that and make any change in a new PR, so as not to delay this one.

Additionally, while making the changes I saw that when using the wizard, going to the wallets screen and then pressing the back button of the phone, the user was returned to the wizard, which did not feel “right”, so I changed that too. It would also be nice to be able to close the modal windows by pressing the back button, but that is a more complex change, so I left that aside for now.

Senyoret1:
Great, I solved both problems 👍

Senyoret1:
Wow, good catch. I did not see the error because in fact it is really weird: it only happens if the browser window is quite tall (some piece of Angular code removes the `.mat-snack-bar-handset` class from the snackbar container if the window is tall and has any of the widths you indicated, but only if the window is tall). With the last change it should work.

The modal window that triggers a snackbar is the one used for creating or loading a wallet. You can test it trying to create a wallet with a repeated seed.

About the general css rule, I agree, I do not like the current solution that much, but it's not that simple. The position of the snackbar must be different when a modal window is being displayed (because the modal windows appear above the navigation bar), but the snackbar is not placed in the DOM inside the modal windows or inside the general content of the app, but on its own div, above those two elements, so it is not simple to calculate just with css what the correct position is. If you think of something I'm all ears, but what I have in mind at this moment is to create a custom snackbar, which should allow to create a more elegant solution and should also make it easier to customize it (I would like to change the color, to show confirmation messages, and put a button for closing it, for example).

mahansky:
Ok let's leave it as it is (the issue is no longer present, it works correctly), I like the idea of custom snackbar, that could bring some benefits

PR #545, opened 2019-04-18
Add CSP response header

### LINKED ISSUE #537: Use a CSP
The CSP should be similar to the one added in https://github.com/skycoin/skycoin/pull/2228 . This means adding it to the server and finding a solution for builds made to be opened from the local file system.

iketheadore:
Fixes #537 

Changes:
- Update nginx config in docker to add CSP
- Add CSP middleware handler to the electron server

### DIFF
--- a/docker/images/nginx.conf
+++ b/docker/images/nginx.conf
@@ -43,6 +43,8 @@ http
     # ssl_certificate /etc/nginx/ssl/server.crt;
     # ssl_certificate_key /etc/nginx/ssl/server.key;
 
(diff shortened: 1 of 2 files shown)


Senyoret1:
I was checking this PR and I could see the csp header appearing correctly in the browser. However, this csp has the problem of blocking access to the servers of other coins and any custom servers that the user would like to add. If we want to leave those options active, the solution is much more complicated.

In the Electron build we could use Electron itself to ask for confirmation before adding access to another URL. For the version that runs from the local file system I think that a modified version of the server used by the Electron version could be added, also to add exceptions. However, no good ideas comes to my mind for the web version.

I think this is not a priority and, because of that, we could take some time to look for a good solution. That do you think @gz-c ?

Senyoret1:
A temporary solution could be simply to block in the web version the possibility of using a custom node and update the server's CSPs as new coins are added.

gz-c:
I guess we should remove the custom URL option for the deployed/hosted web wallet.

For locally run web wallet, we'd still want the custom URL to allow the user to choose their backend node.  However I do not want to enable nodejs integration in electron to support it.

@Senyoret1 remove the ability to set a custom node for the hosted version, but keep it in the electron version

PR #555, opened 2019-04-30
Code fixes

### LINKED ISSUE #550: Maintain the `last_seed` in wallet struct
The `last_seed` is in the `Address` struct, and each time when need to add a new address, it looks the last address in the wallet to find the `last_seed`, and all addresses have a `last_seed` data. 

We can maintain the `last_seed` in the `Wallet` struct, so we can get it directly from the wallet when need to add new address, also in the future, if we have encryption method added, we only need to maintain the `last_seed` and private keys of addresses in a map, which is what we're doing in the desktop wallet now.

### LINKED ISSUE #553: Error: Cannot read property 'decimalPlaces' of undefined in SendFormComponent.html
To reproduce the error:
Set the `environment.production` as false and create some wallets first, then open the send page, and refresh the wallet. The error will be shown as below:
image

This is caused by the `undefined` of `wallet.hours` in `send-form.component.html` file.  when `environment.production` is false, the app will load wallets from `localStorage`, and if the wallets we saved previously do not have coins, the `hours` and `balance` properties won’t be exist, thus when the code try to run `decimalPlaces` of `hours`, the error will be thrown. 

To fix the error:

replace the code of 
```html
wallet.hours.decimalPlaces(0).toString() | number:'1.0-0' 
```
with 
```html
( wallet.hours ? wallet.hours.decimalPlaces(0).toString() : 0) | number:'1.0-0' }} {{ currentCoin.hoursName }})
```

Senyoret1:
Fixes #550 
Fixes #553 

Changes:
- Now the next seed for creating an address is stored in the `Wallet` object and not in the individual `Address` objects.
- The `Cannot read property 'decimalPlaces' of undefined` error in the form for sending coins was fixed.
- The unit tests have been updated.

### DIFF
--- a/e2e/wallets.po.ts
+++ b/e2e/wallets.po.ts
@@ -154,12 +154,11 @@ export class WalletsPage {
 
   addAddress() {
     return element(by.css('.-btn-plus')).click().then(() => {
-      return browser.wait(() => {
-        const lastRecord = element.all(by.css('.-record')).last();
-        return lastRecord.element(by.css('.address-column')).getText().then((address) => {
-          return address === 'bSp3JvfGiHzumCpdaWT7tGRtejwhLDd2zv';
-        });
-      }, 10000);
+      browser.wait(ExpectedConditions.invisibilityOf(element(by.css('-disabled-span'))), 10000);
+      const lastRecord = element.all(by.css('.-record')).last();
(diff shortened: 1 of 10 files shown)


gz-c:
@iketheadore

PR #559, opened 2019-05-09
Deactivate the custom URLs for the nodes

Senyoret1:
Changes:
- Deactivates the option for changing the node URL in the nodes page. This does not happens if the wallet was compiled for being used from the local file system or Electron.


gz-c:
travis failed?

Senyoret1:
Fixed

PR #564, opened 2019-05-19
Fixes for the option of using custom nodes in the Electron version

Senyoret1:
Changes:
- Now if the user tries to change the URL used to access the node of a coin, while using the Electron version, the user must accept the operation via a modal window. This way Electron can block access to all unknown hosts and only add exceptions when the user confirms them.

- The way in which the version that runs from the local file system works was not affected.

- The `connect-src` part of the  CSP of the the server used by the Electron version was changed from `'self' https://api.coinpaprika.com https://node.skycoin.net` to simply `*`. This is because the Electron code already blocks all connections to unknown hosts and the previous CSP would not allow the Electron code to add exeptions for custom nodes. In fact, the option to modify the URL stopped working after the CSP was added.

- Now the option for setting custom URLs is active in the test build.

### DIFF
--- a/electron/src/electron-main.js
+++ b/electron/src/electron-main.js
@@ -236,8 +236,9 @@ function createWindow() {
   });
 }
 
+const customNodeUrlsFilePath = path.join(app.getPath('appData'), './Skycoin');
 // File with the info about the custom node URLs added by the user.
-const customNodeUrlsFile = './nodes.snd'
+const customNodeUrlsFile = path.join(customNodeUrlsFilePath, './custom-nodes.snd');
 
 // Load, from customNodeUrlsFile, the custom node URLs added by the user and add them to the URL whitelist.
 ipcMain.on('loadNodeUrlsSync', (event) => {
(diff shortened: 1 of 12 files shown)


electron/src/electron-main.js
```diff
@@ -229,14 +236,100 @@ function createWindow() {
   });
 }
 
-// Gets the URLs of the nodes of all the coins and adds the hosts to allowedNodes, so that it is
-// possible to connect with them.
-ipcMain.on('setNodesUrls', (event, list) => {
-  allowedNodes = new Map();
-  list.forEach(value => allowedNodes.set(url.parse(value).host, true));
+// File with the info about the custom node URLs added by the user.
+const customNodeUrlsFile = './nodes.snd'
```
gz-c: Does electron have a global storage location? Or, at least shouldn't we store this in one of the standard app config locations?

Senyoret1: Electron does not have a default location for saving files, but it is possible to get the path of the system folders. Here is a list of the options Electron has: https://github.com/electron/electron/blob/master/docs/api/app.md#appgetpathname

gz-c: Can we use localStorage?

gz-c: otherwise use `appData` or `userData` from that docs link

Senyoret1: `localStorage` can only be used inside the app window, not the main Electron code, and, as this is a security feature similar to the CSP, I would not like to involve the app code. I modified the code so that the file is now saved inside `appData`. I also solved a small bug that made some URLs to continue to be whitelisted during the session after being replaced by the user.

iketheadore:
LGTM

iketheadore (approved):

PR #570, opened 2019-06-20
Use the wasm version of the skycoin-lite lib

Senyoret1:
Changes:
- The GopherJS version of skycoin-lite was replaced with the wasm version.

- The tests have been updated.

- The code now is much better at checking if the library was loaded successfully.

- The compiled skycoin-lite wasm file was added to the repo, but the idea is to load it dinamically as a dependency after merging https://github.com/skycoin/skycoin-lite/pull/32

- It does not work in Android, but apparently the problem will be solved in Go v1.13 (https://go-review.googlesource.com/c/go/+/170950/ ) and the GopherJS version was too slow to be usable anyway


Senyoret1:
The last commit includes an updated version of the wasm lib and the changes needed for making the tests to run. However, the PR for adding the wasm version to Skycoin-lite is still pending, so more changes could be needed later.

gz-c:
merge conflicts

PR #571, opened 2019-07-02
Fix electron wallet cannot open explorer in browser

iketheadore:
The electron wallet will open a terminal when clicking the `Skycoin
Explorer`.

Changes:
- Run `sehll.openExternal()` like we did in skycoin electron wallet to
open the url in default browser when `target = '_blank'`.
- Add `.dockerignore` to reduce the context size we send to docker daemon for building the docker
image.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+node_modules
+electron
+e2e
+dist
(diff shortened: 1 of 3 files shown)


Senyoret1 (approved):
:+1:

PR #581, opened 2019-07-19
Fix problems with the CSP

### LINKED ISSUE #580: Content Security Policy error
The latest code in develop branch will have the following error in console when running the electron build and stuck on the splash page.

> Refused to execute inline script because it violates the following Content Security Policy directive: "default-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-Mtjjozjx7XoFHNx2deDKw/8NpqxBENxnOYvY+com/6Q='), or a nonce ('nonce-...') is required to enable inline execution. Note also that 'script-src' was not explicitly set, so 'default-src' is used as a fallback.

Senyoret1:
Fixes #580 

Changes:
- The inline script that was added to `index.html` was moved to a file, to stop violating the CSP.

- The `worker-src` rule was removed from the CSP, as the web worker used by the app was replaced by a wasm file, so the rule is not longer needed.

- The `script-src` rule was added to the CSP, because the previous value (inherited from `default-src`) prevented the wasm file from being used in Safari and Chrome (affecting the Electron version). The problem is known by the involved parties, but apparently there is no better solution for now. There is more info about the problem in https://github.com/WebAssembly/content-security-policy/blob/master/proposals/CSP.md

### DIFF
--- a/docker/images/nginx.conf
+++ b/docker/images/nginx.conf
@@ -43,7 +43,7 @@ http
     # ssl_certificate /etc/nginx/ssl/server.crt;
     # ssl_certificate_key /etc/nginx/ssl/server.key;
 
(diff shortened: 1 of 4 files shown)


iketheadore (approved):
👍

PR #583, opened 2019-07-23
Run npm install with --unsafe-perm in Dockerfile

iketheadore:
Changes:

Npm install in the first stage of docker building will report error of:

```
npm WARN lifecycle desktopwallet@0.0.0~prepare: cannot run in wd desktopwallet@0.0.0 npm run snyk-protect (wd=/usr/src/skycoin-web)

```

which will make snyk-protect fail to do the patches.

### DIFF
--- a/docker/images/Dockerfile
+++ b/docker/images/Dockerfile
@@ -5,7 +5,7 @@ COPY . /usr/src/skycoin-web
 WORKDIR /usr/src/skycoin-web
 
 RUN npm install -g @angular/cli \
-    && npm install \
+    && npm install --unsafe-perm \
     && npm run build
 
 FROM nginx:alpine


gz-c:
What does --unsafe-perm do?

Senyoret1:
That flag is for when the npm installThat flag is for when `npm install` is called by the root user or with `sudo`. In that case, npm normally uses a restricted user when running any installation script included with any dependency, but with the flag npm stops doing that and simply continues using the user that called `npm install`.

Probably the restricted user does not have write permission and that makes Snyk unable to apply the patches.

Another option would be simply not to run snyk-protect, as at this time it just installs 2 patches for dev dependencies, which are only used for automatic testing. However, additional changes could be needed in the future if we add a Snyk patch for a runtime dependency.

PR #584, opened 2019-07-27
Add compatibility with the hw wallet

Senyoret1:
Changes:
- This PR adds compatibility with the hw wallet. The compatibility is basically at the same state that the one in the desktop wallet.


Senyoret1:
The lastest commit includes the changes made in https://github.com/skycoin/skycoin/pull/2474

Senyoret1:
The lastest commit shows a warning when changing the name of a hw wallet, to make the user know that any non-standard English character may not be correctly displayed in the hw wallet.

